PR #10, opened 2019-08-08
Homepage: Use trendy fixed background

carlmjohnson:


carlmjohnson:
Does not work on mobile Safari.

PR #70, opened 2020-02-17
JS: Experimental use of Alpine.JS

carlmjohnson:


carlmjohnson:
"Learnings": Putting stuff into the DOM feels weird after using Vue but seems to be the way to use Alpine.

PR #80, opened 2020-04-16
Right size images

carlmjohnson:


carlmjohnson:
Yikes, adds a second to load.

PR #83, opened 2020-04-20
Move readmore to module

carlmjohnson:


carlmjohnson:
Parcel doesn’t really support being in a module.

PR #92, opened 2020-04-29
Giving Tuesday promos

carlmjohnson:


carlmjohnson:
Todo: hoist analytics calculations.

PR #153, opened 2020-09-21
Layout: Only load one tote promo

carlmjohnson:


carlmjohnson:
I don't want the layout shift. It's not worth the savings.

PR #154, opened 2020-09-22
Add related links

carlmjohnson:


assets/css/featured-tailwind.css
```diff
@@ -63,6 +61,15 @@ html {
   @apply list-decimal;
 }
 
+@supports (shape-outside: border-box){
```
carlmjohnson, line 64: Not compatible with `@responsive`.

PR #170, opened 2020-12-10
JS: Import ga-lite without Babel

carlmjohnson:


carlmjohnson:
Saves 3kb. Seems to work in IE11

PR #194, opened 2021-06-16
Alpine 3

carlmjohnson:


carlmjohnson:
https://github.com/tdewolff/minify/issues/422

PR #235, opened 2021-11-12
Static: Use preload for CSS

carlmjohnson:


carlmjohnson:
Neat but has a huge FOUC problem for no real gain.

PR #284, opened 2022-07-05
TND New About Section

regisphilibert:
Hi @carlmjohnson 

Here’s what you need to know:  

We’ve tried and componentized our work as much as possible so everything lives under the `about-tnd` content directory. This means we're using our own content type, and its layouts lives under `layouts/about-tnd`. I guess the work would now revolves around making those templates the base for your own about section, and moving the content files in it if the copy appears final to you.

We are using a block solution to populate the about landing page, as well as its children even though those only use one block printed after the body content.

We are now using TW globally. After your initial reservations, I carefully studied Bulma and realized their only conflicting utility classes are of spacing. (`p-4`, `my-3` etc…). But as they behave exactly as TW (1 unit being .25 rem) this does not compromise the integrity of your current design system. I simply disabled Burma’s spacing utility classes, so now any spacing reference is handled by Tailwind without any visual change.

https://github.com/spotlightpa/poor-richard/blob/762ebc5f1a0d0c33987e3f353deaf8fb8459b9ff/assets/css/bulma.scss#L87

Feel free to reach out if you need anything else from us. We could also set up a call and walk you through the logic.

Thanks.


carlmjohnson:
Thanks. We're working on launching the State College bureau this week and then I'm going on vacation, but I'll try to land it as soon as I can after that, probably by the end of July. I'll let you know if I have questions.

carlmjohnson:
@budparr What's the source of the donate.svg?

carlmjohnson:
Thanks, looks good.

carlmjohnson:
Merged in #292

PR #343, opened 2023-01-04
Jr/tw newsletter archives

jeffrummel:
Updates newsletter archives to tailwind


carlmjohnson (changes requested):
It has the wrong footer now.

layouts/_default/newsletters-archive.html
```diff
@@ -25,40 +29,40 @@
...
+      <nav class="mb-16 mx-auto max-w-screen-lg">
         {{ range $pages.GroupByDate "January 2006" }}
           <div class="tile is-parent">
-            <h2 class="title">{{ .Key }}</h2>
+            <h2 class=" text-3xl font-extrabold font-raleway mt-16 mb-9">
+              {{ .Key }}
+            </h2>
           </div>
-          <div class="tile is-wrapped">
+          <div class="mt-4 grid gap-12 sm:grid-cols-2 md:grid-cols-3">
```
carlmjohnson, line 65: Don't use the tile/list partial here. I want the tile stuff to be pure Bulma so we can drop the whole directory when Bulma is removed. You can use the tw/list-3-col-story.html partial instead. See how _default/list.html and _default/terms.html do it.

layouts/partials/component/image-block.html
```diff
@@ -80,7 +80,7 @@
     {{- end -}}
     {{ with $credit }}
       <figcaption
-        class="is-clearfix is-size-7 is-uppercase has-text-grey-light"
+        class="is-clearfix is-size-7 is-uppercase has-text-grey-light mt-1 w-full px-1 text-right font-sans text-xs uppercase leading-tight text-credit"
```
carlmjohnson, line 83: See above. We can just use the tw stuff, so this doesn't need to change.

layouts/shortcodes/newsletter/investigator-form.html
```diff
@@ -10,14 +10,18 @@
     data-ga-category="newsletters:investigator-full"
```
carlmjohnson, line 10: Prettier only does the Tailwind class sorting stuff to a whitelisted set of files. Add `"**/shortcodes/newsletter/*.html",` to prettier.config.js and it will start sorting this file too.

carlmjohnson: The three shortcodes are pretty similar. It might be time to extract them into a common base partial and just pass a dict of differences from each file.

layouts/partials/component/image-block.html
carlmjohnson: mt-1 has different size in Bulma than Tailwind, so it ends up looking a little different if you try to write one file that works in both. For the nav I just used normal BEM and then duplicated the styles in bulma.css and tw.css.

carlmjohnson:
One thing to know about the paginator, is Hugo has some weird issue where because paginating creates subpages at /page/2 etc you can only trigger it once per page. If you trigger it wrong the first time, you can't fix it lower down on the page. Since I want to have (page 2) in the title on /page/2/, we have to jump through some hoops to make sure the paginating is already set up correctly in the title tag. See these: https://github.com/spotlightpa/poor-richard/blob/master/layouts/partials/helper/paginator.html https://discourse.gohugo.io/t/control-pagination-and-page-collections-from-baseof-html/37643/7

layouts/_default/newsletters-archive.html
carlmjohnson: This layout has choppy loading because it isn't getting the aspect ratio hint for images, I guess because it's relying on the Bulma aspect-ratio helper, so switching to list-3-col-story will be good for fixing that and getting the subgrid alignment automatically.

carlmjohnson:
You know what? I just removed the custom layout on the _index.md files, and it dropped back to list.html, and that looks good now that the shortcode is updated. We can just delete newsletters-archive.html.

layouts/shortcodes/newsletter/investigator-form.html
jeffrummel: Updated
https://github.com/spotlightpa/poor-richard/pull/343/commits/b2f329f83a5d2fe01c6f8fd9fa8141f61d001ed9

jeffrummel:
Removing the newsletter-archives.html and tiles/list.html also removes the [separation of posts by] month(https://www.spotlightpa.org/newsletters/investigator/#archives)

Are you sure we want to remove that too?

carlmjohnson:
This worked:

_index.md

```toml
layout = "news.html"
sort-by = "month"
```

helper/paginator.html

```
  {{ else if eq $sort "month" }}
    {{ $p = .Paginate (.Pages.GroupByDate "January 2006") $size }}
```

_default/news.html

```html
{{ define "extra-main" }}
  <section class="bg-white px-4 pb-12 pt-6" id="archives">
```

image

layouts/shortcodes/newsletter/papost-form.html
```diff
@@ -10,52 +10,60 @@
...
+          class="input border font-sans w-full border-s-4 rounded p-1"
           type="text"
           name="LNAME"
           required
         />
       </p>
     </div>
-    <div class="field">
+    <div class="field mb-3">
       <label class="label" for="{{ $idEmail }}">Email Address</label>
```
carlmjohnson, line 49: This one was missed. 

image

jeffrummel:
> This worked:
> 
> _index.md
> 
> ```toml
> layout = "news.html"
> sort-by = "month"
> ```
> 
> helper/paginator.html
> 
> ```
>   {{ else if eq $sort "month" }}
>     {{ $p = .Paginate (.Pages.GroupByDate "January 2006") $size }}
> ```
> 
> _default/news.html
> 
> ```
> {{ define "extra-main" }}
>   <section class="bg-white px-4 pb-12 pt-6" id="archives">
> ```
> 
> image

These changes did not work for me
Screenshot 2023-01-04 at 4 25 23 PM

jeffrummel:
Nevermind. For whatever reason `layout = "news.html"` doesn't work but `layout = "news"` does

carlmjohnson (commented):
Okay, delete newsletters-archive.html and revert the changes to image-block.html and tile/list.html, and I think this will be ready to go.

PR #351, opened 2023-01-27
Add Privacy Policy & ToS

carlmjohnson:


carlmjohnson:
CC: @jeffrummel. Not sure if this will be affected by what's on your branch if the about pages get redone. One of the pages has a table.

jeffrummel:
thanks for the flag. My branches should be okay

PR #353, opened 2023-01-31
Layout: Redo events page in Tailwind

carlmjohnson:


carlmjohnson:
Screenshot 2023-01-31 at 1 31 49 PM

What it looks like when there's an upcoming event

carlmjohnson:
@jeffrummel I redid the events page. As part of it, I noticed the content well is repeated a bunch, so I pulled it out into its own partial.

PR #355, opened 2023-02-01
Jr/new donation

jeffrummel:
Hey @carlmjohnson

We could probably simplify the donation shortcode and partial files using a helper, but I haven't fully wrapped my head around how Hugo passes parameters from one template to the next. Please let me know if you have ideas or thoughts on this.

I can't imagine we'll use the shortcode very often, but I figured it might be nice to have it in our toolbox, just in case.

**Replacement for  tw/promo-footer**
```
{{ define "before-footer" }}
  {{ partialCached "tw/donate-prefill-lg.html" (dict
    "donateHed" "Donate to State College"
    "donateCta" "Give Today"
    "donateCampaign" "7015G0000013pUYQAY"
    "donateTheme" "centrecounty"
    "donateAmounts" (slice "12" "27" "50" "100")
    )
  }}
{{ end }}
```
**New Shortcode:**
```
{{<featured/donate-prefill-lg
  donateHed="Here We Go!"
  donateGaCategory="donate-prefill-lg"
  donateCampaign=""
  donateCta="Donate Today"
  donateAmounts="12,15,50,100"
>}}
```

Thanks


carlmjohnson:
I think the way to do it is to have the shortcode just pass its arguments to a normal partial, so all the HTML is in the partial and all the shortcode is doing is a quick unpack and call.

layouts/partials/tw/donate-prefill-lg.html
```diff
@@ -0,0 +1,209 @@
...
+{{ $btnClass := ` mr-4 flex-[0_1_30%] outline-none min-w-[180px] inline-block rounded py-3 px-6 text-white font-sans font-black text-lg tracking-wide ring ring-transparent` }}
+{{ $activeClass := `bg-red-6 text-white hover:bg-red-5 focus:ring-red-7 active:ring-red-9` }}
+{{ $inactiveClass := `bg-s-2 text-s-9 hover:bg-s-3 focus:ring-s-7 active:ring-s-9` }}
+{{ $radioClass  :=  `rounded-full border-s-1 appearance-none mr-4 order-first w-6 h-6 checked:bg-black shadow-[0_0_0_3px_rgba(255,255,255,1)]` }}
+
+{{ $donateAmount := split $donateAmounts "," }}
+{{ $donateFirst := index $donateAmounts 1 }}
+{{ $donateDefault := int $donateFirst }}
+<div
+  id="donate-leader"
```
carlmjohnson, line 20: Do we need this id? Probably we'll only have one of these on the page, but I don't like that it limits us if we wanted more.

layouts/shortcodes/featured/donate-prefill-lg.html
```diff
@@ -0,0 +1,20 @@
+{{ $wrapperClass := .Get "wrapperClass" | default "w-[calc(100vw - (100vw - 100%))] my-10 -mx-[50%]"}}
```
carlmjohnson, line 1: Does this wrapper class work? ISTM, that Tailwind need the spaces to be underscores.

PR #357, opened 2023-02-02
Jr/tw staff

jeffrummel:
@carlmjohnson I've converted Staff to Tailwind

### DIFF
--- a/layouts/_default/staff-page.html
+++ b/layouts/_default/staff-page.html
@@ -125,7 +125,6 @@
     id="{{ $author.slug }}"
     class=" mb-8 grid grid-cols-5 grid-rows-[248px_3fr] gap-5 rounded bg-white p-5 shadow-md lg:flex-[0_1_calc(50%_-_1rem)] font-raleway"
   >
-
     {{ if $author.thumb }}
       <img
         src="{{ $author.thumb }}"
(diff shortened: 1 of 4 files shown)


layouts/_default/staff-page.html
```diff
@@ -1,3 +1,77 @@
...
+    <div
+      class="mx-auto max-w-[1248px] items-center md:min-h-[110px] gap-y-3 gap-x-5 md:grid md:grid-cols-12"
+    >
+      <h1
+        class="uppercase font-serif text-[45px] leading-none col-span-5 col-start-2 row-span-1 row-start-1 font-black"
+      >
+        {{ .Title }}
+      </h1>
+      <div
+        class="font-raleway col-span-5 col-start-7 row-start-1 text-xl font-medium mt-3 md:mt-0"
```
carlmjohnson, line 41: The static-tw.html header doesn't include raleway. If you want to use it, you have to do what about-statecollege.html does:

```
{{ define "static-head" }}
  {{ partial "header/css-tw-rale.html" . }}
  {{ partialCached "header/noscript.html" . }}


  <link
    rel="modulepreload"
    href="{{ partialCached `helper/get-enhancements` . }}"
  />

  {{ partial "header/favicons.html" . }}
  {{ partial "header/feeds.html" . }}
{{ end }}
```

Probably that should be pulled into a partial called header/static-tw-with-raleway.html or something.

carlmjohnson (commented):
Very exciting overhaul.

layouts/_default/staff-page.html
```diff
@@ -1,3 +1,77 @@
...
+  </header>
+{{ end }}
+
+{{ define "main" }}
+  <header class="bg-black py-14 px-3 text-white">
+    <div
+      class="mx-auto max-w-[1248px] items-center md:min-h-[110px] gap-y-3 gap-x-5 md:grid md:grid-cols-12"
+    >
+      <h1
+        class="uppercase font-serif text-[45px] leading-none col-span-5 col-start-2 row-span-1 row-start-1 font-black"
```
carlmjohnson, line 36: The classes aren't being sorted by prettier because it thinks this is a Bulma page. Add _default/staff-page.html to line 23 of prettier.config.js.

Also `font-black` is an ultrabold. I find it confusing, because you would think it's just the same as `text-black`. Maybe it would be more clear if we used `font-[900]` instead? Up to you.

jeffrummel: I've gotten used to setting 900 weight with font-black ATP so I feel good about keeping it.

jeffrummel:
@carlmjohnson should I remove Nick Beiling and Sarah Hutchins before merging?

carlmjohnson:
You can just merge if it's ready, but if you want to fix them, just change their toml files in data/authors from active = true to active = false. It can also be a follow up commit when one of us gets around to it.

carlmjohnson (approved):

PR #359, opened 2023-02-06
Jr/staff rwd fix

jeffrummel:


carlmjohnson:
The small breakpoint is 640px. I did a quick check, and I think it looks okay with sm as the trigger instead of md:

image

PR #365, opened 2023-02-21
statecollege/list.html: Add data-sf-theme to make /donate/ links correct

carlmjohnson:


jeffrummel (commented):
body attributes should probably be added to `layouts/_default/transparency-tracker.html` too.

PR #367, opened 2023-02-24
Jr/ribbon modal

jeffrummel:
New newsletter-modal design


layouts/partials/tw/modal-ribbon.html
```diff
@@ -0,0 +1,166 @@
...
+  <div
+    class="absolute inset-0 backdrop-blur-sm backdrop-filter"
+    x-transition.opacity
+  ></div>
+  <div
+    class="absolute inset-0 flex items-center overflow-y-auto"
+    x-show="isOpen"
+    x-transition
+  >
+    <div class="w-full border-b-2 border-t-2 bg-[{{ $actionColor }}] font-sans">
```
carlmjohnson, line 33: If you don't say exactly bg-[whatever] in the HTML, the CSS purge script can accidentally purge this class out. Usually, I give up and set a style. Another solution is to do `style="--bg-color: {{ $actionColor }}" class="bg-[--bg-color]"`.

jeffrummel: Thanks that worked!

jeffrummel:
@carlmjohnson I tried loading `{{ $newsletterAction := $.Site.Param `newsletters-action`}}` in a number of different places to get it to work without any luck, so LMK if you have a better spot than this [layouts/_default/index.html](https://github.com/spotlightpa/poor-richard/pull/367/commits/e7908b1ef990d6f051eea9ae2b31a5bc65189dd5)

carlmjohnson:
In Hugo, local variables start with $, the current context starts with dot, and functions start with a lowercase letter. `$.Site` means "look at .Site on the root context for this block". It lets you move up a level from the context of a `range` or `with` block, but you're still stuck in the current template context. Because `.Site` is so useful, Hugo also has a verb that provides it, and unlike the context, you don't have to pass it around everywhere. I generally try to just always use `site` instead of `.Site` because it doesn't need a context. Try `$newsletterAction := site.Param "newsletters-action"` and see if that fixes it.

layouts/partials/tw/modal-ribbon.html
```diff
@@ -166,8 +166,8 @@ <h2 class=" text-2xl font-extrabold md:text-3xl lg:text-4xl">
           <label>
             <input
               type="checkbox"
-              class=" mr-3 h-4 w-4 rounded border border-s-4 p-1 hover:accent-yellow-4"
-              style="accent-color: {{ $actionColor }}"
+              class=" mr-3 h-4 w-4 rounded border border-black p-1 hover:accent-yellow-4"
+              style="accent-color: black"
```
carlmjohnson, line 170: If it's not a variable anymore, you can use the class accent-black.

layouts/_default/author.html
```diff
@@ -257,7 +257,10 @@ <h3 class="my-2">
       {{ partialCached "tw/modal-sticky.html" . }}
     {{ end }}
     {{ if .Param "newsletter-active" }}
-      {{ partialCached "tw/modal-newsletter.html" . }}
+      {{ partialCached "tw/modal-ribbon.html" ( dict
```
carlmjohnson, line 260: If you use partialCached instead of plain partial, does it ever get confused about which modal is which?

jeffrummel: Good call. We're now just loading partial for the modal ribbon. https://github.com/spotlightpa/poor-richard/pull/367/commits/7733d03fd4527a3d242088c275721750ae4d8e0c

carlmjohnson (approved):

PR #372, opened 2023-03-02
Layouts: Remove unused files

carlmjohnson:


carlmjohnson:
`hugo --printUnusedTemplates` prints out the unused templates. It has false positives because of the ad slots that are turned off ATM, but it gives a rough guide.

jeffrummel (approved):
this looks good to me

PR #380, opened 2023-03-09
Spacing between hed and img in river

jeffrummel:


carlmjohnson:
Does this move the hed when it's not wrapped? What about adding a gap-y-5 on the container instead?

jeffrummel:
It does not move the hed when wrapped. I can change to gap if that's
your preference.

On March 9, 2023, GitHub ***@***.***> wrote:
> Does this move the hed when it's not wrapped? What about adding a gap-
> y-5 on the container instead?
>

—
Reply to this email directly, view it on GitHub
<https://github.com/spotlightpa/poor-richard/pull/380#issuecomment-
1462200114>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAGH3KTC5I5QJVFCMKQKAYTW3HVSFANCNFSM6AAAAAAVVGAKOE>.
You are receiving this because you authored the thread.Message ID:
***@***.***>

carlmjohnson:
I think just merge it as is. If we need to tweak it again, it can be set then.

PR #388, opened 2023-03-30
Yarn upgrades

carlmjohnson:


carlmjohnson:
It appears to be linked to https://github.com/tailwindlabs/tailwindcss/pull/10672. I'm going to avoid upgrading Tailwind until we know more.

PR #397, opened 2023-04-13
11k+ links updated

jeffrummel:


carlmjohnson:
I don't want us to have to do this again. Can you just make them all /donate/?

carlmjohnson (approved):
Seems good. Have Joanna poke around and if she's happy we can merge it.

PR #402, opened 2023-04-25
New footer

jeffrummel:


carlmjohnson:
image

Give the input a border radius, so they are both rounded. 

image

Sizing here needs to be consistent. 

image

This goes back to some questions on the style guide PR, but I think this should be all caps and text-yellow-9.

image

There should be a space before the start of the hyperlink.

image

The page is overflowing at tablet width and below.

I think the red support button needs to come down at tablet, and the text should shrink so it's not longer than 3 lines long.

layouts/partials/tw/footer.html
```diff
@@ -1,59 +1,180 @@
...
+        x-cloak
+        x-show="hasFocused"
+        x-transition:enter="transition ease-in-out duration-300"
+        x-transition:enter-start="opacity-0"
+        x-transition:enter-end="opacity-100"
+        x-transition:leave="transition ease-in duration-300"
+        x-transition:leave-start="opacity-100"
+        x-transition:leave-end="opacity-0"
+      >
+        {{ $checkClass := `accent-orange` }}
```
carlmjohnson, line 100: Drop this line.

PR #356, opened 2023-02-01
Jr/transparency tracker

jeffrummel:

PR #411, opened 2023-05-08
Show/Hide byline on thumb-v-item

jeffrummel:
Removes list-3-col-story-nobyline referenced here: https://github.com/spotlightpa/poor-richard/pull/356/files/a2152b7ec504c792934fc4f4994bc5bb4c9e48e4#r1110245650


layouts/partials/tw/thumb-v-item.html
```diff
@@ -1,3 +1,6 @@
+{{ $showByline := .showByline | default "true" }}
```
carlmjohnson, line 1: Make it `default true` with no quotes around `true`.

carlmjohnson: image

It looks like it's not working. I think the problem is that the default kicks in even when you explicitly set it to false because it replaces falsey values. To fix it, make the options work the other way around and have `{{ $hideByline := .hideByline | default false }}` instead of .showByline and that should work.

carlmjohnson: Looks like the undefined is passing through correctly now, but I'd still prefer to have `true` with no quotes instead of `"true"` as the default.

carlmjohnson (changes requested):
Good to go once the quotes are off "true".

carlmjohnson (approved):

PR #413, opened 2023-05-08
Spring Cleaning - removes unused files/directories

jeffrummel:


layouts/partials/tw/image-block.html
```diff
@@ -7,6 +7,7 @@
 {{ $eager := .eager | default false }}
 {{ $gravity := .gravity }}
 {{ $rounded := cond (.rounded|not|not) "rounded" "" }}
+{{ $shadow := cond (.shadow|not|not) "shadow" "" }}
```
carlmjohnson, line 10: Note that the "not not" is a workaround for this Hugo bug: https://github.com/gohugoio/hugo/issues/5792 https://github.com/gohugoio/hugo/issues/9160

carlmjohnson (approved):

PR #415, opened 2023-05-08
Layout: Use global page.Store for SVGs

carlmjohnson:
This should solve the problem of missing SVGs on the page and optimize pages that don't use all the SVGs. From now on instead of `<use href="#x">` we should use a partial that will add the SVG to the page global list of SVGs used.


jeffrummel (approved):
This is great

PR #416, opened 2023-05-09
Page global SVGs attempt 2

carlmjohnson:
I'm still not sure why this exploded before. I think maybe some stray punctuation broke the minifier and that broke the rest of the page, but I'm still trying to prove that experimentally.


carlmjohnson:
Reworked into #417.

PR #422, opened 2023-05-31
Jr/new header

jeffrummel:


carlmjohnson:
image

There's a little 1px misalignment of the items on my desktop.

image

And the menu overlaps on my phone.

jeffrummel:
> image
> 
> There's a little 1px misalignment of the items on my desktop.
> image
> 
> And the menu overlaps on my phone.

These are fixed

carlmjohnson:
My suggestion to work around the CSS purge issue is instead of using `:class`, see if you can use `x-show` with `x-transition`, and then you'll get a nice pop-in animation: https://alpinejs.dev/directives/transition. You should also add `x-cloak`, so the menu starts out invisible before the JS kicks in.

layouts/partials/tw/footer.html
```diff
@@ -17,26 +17,36 @@
...
+           sm:flex sm:w-32 sm:justify-center md:w-52 md:px-7 md:py-4
           "
         href="/donate"
       >
-        <svg
-          class="mr-2 h-4 w-4 fill-current sm:h-6 sm:w-6 md:mr-6 md:h-9 md:w-9"
-          aria-hidden="true"
+        <span
+          class="
+          absolute inset-0 z-0 bg-red [clip-path:_polygon(0_0,_0%_100%,_15%_50%)] group-hover:bg-blue
```
carlmjohnson, line 32: `[clip-path:_polygon(0_0,_0%_100%,_15%_50%)]` seems like it might be useful enough to add to tw.css as `.clip-circle` something.

layouts/partials/tw/image-rotator.html
```diff
@@ -5,7 +5,7 @@
 
 <span class="mx-auto block max-w-full" style="width: {{ .width }}px;">
   <span
-    class="{{ $bgClass }} relative block h-0 w-full overflow-hidden"
+    class="{{ $bgClass }} relative block h-0 w-full max-w-screen-xl mx-auto overflow-hidden"
```
carlmjohnson, line 8: I haven't looked into this, so tell me if I'm wrong: shouldn't the max-w-screen-xl go on the container, because otherwise padding can be for wrong size? It might be time to rewrite this to use aspect-ratio to make the code more clear and cut out the middle span.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,268 @@
-<nav
+<div
   id="top"
   x-data="{ isOpen: false }"
-  class="site-nav"
```
carlmjohnson, line 4: Let's delete the old .site-nav classes from tw.css.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,268 @@
...
+    </a>
+  </h1>
+  <a
+    data-sf-campaign="701Dn000000YgogIAC"
+    data-ga-category="donate"
+    class="group relative inline-flex items-center justify-around overflow-hidden rounded-md bg-gradient-to-r from-red-7 to-red p-2 hover:from-blue-7 hover:to-blue sm:flex sm:w-32 sm:justify-center lg:w-52 lg:px-7 lg:py-4"
+    href="/donate"
+  >
+    <span
+      class="absolute inset-0 z-0 bg-red [clip-path:_polygon(0_0,_0%_100%,_15%_50%)] group-hover:bg-blue"
```
carlmjohnson, line 31: Here's that .clip-circle again.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,268 @@
...
+  <button
+    class="
+      no-js:hidden group z-20 order-1 flex flex-[0_1_70px] flex-row items-center gap-1 self-stretch text-sm font-bold leading-none tracking-wide
+    "
+    :class="toggle ? 'sm:mr-3 sm:-ml-3' : 'mx-0'"
+    type="button"
+    x-ref="button"
+    aria-label="Site Menu"
+    :aria-expanded="toggle"
+    :aria-controls="$id('dropdown-button')"
```
carlmjohnson, line 88: I think this doesn't work because it doesn't have a corresponding `x-id="['dropdown-button']"` on the thing it opens. See https://alpinejs.dev/directives/id and https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,268 @@
...
-      <div class="site-nav:mobile">
-        <div
-          x-data="{ isOpen: false }"
-          x-on:keydown.escape.prevent.stop="isOpen = false; $refs.button.focus()"
-          x-id="['dropdown-button']"
-          style="position: relative;"
+          class="
+            sm:h block w-full flex-1 border-b py-4 font-bold uppercase hover:underline hover:underline-offset-2 sm:flex sm:w-full sm:items-center sm:capitalize
+          "
+          :class="toggle ? 'sm:py-0 sm:h-11 md:justify-start' : 'md:py-0 md:justify-end'"
```
carlmjohnson, line 150: I think it would be simpler to just duplicate the top level items inside and out of the pop up menu than to have the inner items changing classes based on the toggle.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,268 @@
...
+  h-auto w-0
+  w-0 w-auto translate-y-1.5
+  rotate-45 overflow-hidden rounded border-s-5 px-3 px-4 px-5 pb-10 pt-12 opacity-0 max-sm:absolute max-sm:border sm:-right-16 sm:-right-16 sm:-top-6 sm:top-52
+  sm:-ml-3 sm:mr-3 sm:block sm:block sm:block sm:h-11 sm:h-11 sm:h-40 sm:w-0 sm:w-0 sm:w-0 sm:w-48 sm:w-48 sm:w-48 sm:w-48 sm:overflow-hidden sm:overflow-hidden sm:overflow-hidden sm:overflow-visible sm:overflow-visible sm:overflow-visible sm:overflow-visible sm:rounded-l
+  sm:rounded-br sm:rounded-tr sm:border-b sm:border-b sm:border-l
+  sm:border-r sm:border-r sm:border-t sm:border-t sm:border-s-5
+  sm:border-s-5 sm:border-s-5 sm:bg-white sm:bg-white sm:bg-white sm:px-16 sm:px-3 sm:py-0 sm:py-0 sm:pb-0 sm:pb-10 sm:pb-3 sm:pb-[35px] sm:pl-4 sm:pl-4 sm:pl-8 sm:pr-4 sm:pr-8
+  sm:pr-8 sm:pt-0 sm:pt-2 sm:pt-2 sm:pt-6 md:absolute md:-top-4 md:-top-8 md:right-1 md:top-0 md:hidden md:h-auto md:w-0 md:w-full md:justify-start md:justify-start md:justify-end md:justify-end
+  md:overflow-visible md:py-0 md:py-0 md:pt-[15px] md:pt-[15px]
+"
```
carlmjohnson, line 259: Add a `<!-- comment -->` explaining why we need the invisible spans for CSS purge.

carlmjohnson:
I can't tab to the more button because the outlines are hidden.

The menu should probably close if it detects an escape keydown or a click outside.

carlmjohnson:
It can be a follow up PR, but we should redo the about page to use the new nav.

carlmjohnson:
If you disable JS, the screen is screwed up.

Screenshot 2023-06-01 at 4 29 16 PM

This matters because there's a really quick flash during page load before the JS kicks in and it loads that, and then it goes over the final view. The solution is to make sure the untoggled state is in all the plain `class=` attributes, and reserve `:class` for toggled on classes.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,269 @@
...
+        ></span>
+      </div>
+    </div>
+  </button>
+  <ul
+    class="
+      -top-2 right-1 z-10 ml-1 box-content flex-wrap gap-2 bg-white shadow sm:relative sm:top-auto sm:flex sm:h-auto sm:w-auto sm:max-w-4xl sm:flex-1 sm:flex-nowrap sm:items-center sm:justify-evenly sm:rounded-none sm:border-none sm:bg-transparent sm:bg-none sm:px-16 sm:shadow-none sm:before:content-[] md:overflow-visible md:px-[6.75rem]
+    "
+    :class="toggle ? 'max-sm:absolute max-sm:border rounded border-s-5 flex w-auto h-auto px-4 pt-14 pb-10 sm:pt-0 sm:pb-0 sm:overflow-visible' : 'overflow-hidden w-0'"
+    x-on="toggle"
```
carlmjohnson, line 124: How does this work? My understanding is that x-on adds an event listener, but toggle is boolean. Is it just a typo of x-show or am I missing some deeper feature of Alpine?

layouts/partials/tw/nav.html
carlmjohnson: I guess the tricky part of just having a second set of menu items is that you want the button animation, and that requires the button to be visible before and after. I think you could do it though if the toggle just transforms the container of the button and shows/hides a second set of items that are next to the button. Those could be x-cloaked and hidden from first view to fix the flicker. Definitely a tricky one though.

layouts/partials/tw/footer.html
jeffrummel: Resolved: https://github.com/spotlightpa/poor-richard/pull/422/commits/c37fc3caf4d4fa73ea3a4f5771c85759067bdee8

layouts/partials/tw/nav.html
jeffrummel: Updated: https://github.com/spotlightpa/poor-richard/pull/422/commits/9e0d40451ab400ba67268e0e18907b7a28bc9df6

layouts/partials/tw/nav.html
jeffrummel: https://github.com/spotlightpa/poor-richard/pull/422/commits/9e0d40451ab400ba67268e0e18907b7a28bc9df6

assets/css/tw.css
```diff
@@ -380,9 +321,16 @@
 .no-js\:block {
   display: none;
 }
+.no-js\:flex {
+  display: flex;
+}
```
carlmjohnson, line 326: To make this work, you also need to add the inverse style to partial/header/noscript.html:

```
  .no-js\:flex {
    display: flex;
  }
```

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,311 @@
...
+        :class="toggle ? 'sm:block sm:right-[-4rem] md:right-[-4rem] sm:-top-6 sm:pt-2 sm:pb-3 sm:pr-8 sm:pl-4 sm:rounded-tr sm:bg-white sm:overflow-visible sm:w-48 sm:h-[auto] md:absolute md:h-auto md:top-[-1rem] md:pt-[15px] box-content sm:border-s-5 sm:border-r sm:border-t' : 'sm:overflow-hidden sm:w-0 sm:h-11'"
+      >
         <a
-          class="site-nav:donate-button"
-          href="/donate/"
-          data-ga-category="donate"
+          class="
+            sm:h block w-full flex-1 border-b py-4 font-bold uppercase hover:underline hover:underline-offset-2 sm:flex sm:w-full sm:items-center sm:capitalize md:py-0
+          "
+          :class="toggle ? 'sm:py-0 sm:h-11 md:justify-start' : 'md:justify-end'"
```
carlmjohnson, line 184: Adding `md:justify-end` to the plain `class` fixes the flicker for me.

carlmjohnson:
The nav on https://deploy-preview-422--poor-richard-spotlightpa.netlify.app/newsletters/ has a white background instead of beige. Is that intentional?

jeffrummel:
> My suggestion to work around the CSS purge issue is instead of using `:class`, see if you can use `x-show` with `x-transition`, and then you'll get a nice pop-in animation: https://alpinejs.dev/directives/transition. You should also add `x-cloak`, so the menu starts out invisible before the JS kicks in.

I originally tried x-show/x-cloak but it was forcing block display when showing and I needed flexbox

layouts/partials/tw/nav.html
jeffrummel: This doesn't work for me because we don't have control over the cascade. `md:justify-end` overwrites `:class= … md:justify-start`

I've tried every combination I can think of to position the class but Alpine cannot override. The only solution I have come up with is an arbitrary class in tw.css https://github.com/spotlightpa/poor-richard/pull/422/commits/25ca397c1d6a1e82640113720c2f9a44b96adf1e

layouts/partials/tw/nav.html
jeffrummel: > How does this work? My understanding is that x-on adds an event listener, but toggle is boolean. Is it just a typo of x-show or am I missing some deeper feature of Alpine?

This is how I understand it as well. I could've named `toggle` anything. x-on is just less opinionated than x-show

layouts/partials/tw/nav.html
jeffrummel: I think this works better from a markup structure and maintenance perspective as a single ul inside the nav. This way we can keep the HTML as structure with CSS adding presentation and JS for DOM manipulation

layouts/partials/tw/image-rotator.html
jeffrummel: This is where it needs to be. But yes, I can change to aspect ratio on another pr

jeffrummel:
> I can't tab to the more button because the outlines are hidden.
> 
> The menu should probably close if it detects an escape keydown or a click outside.

This is fixed

layouts/_default/main.html
```diff
@@ -1,10 +1,14 @@
+{{ define "title" }}
+  <title>{{ .Param "title-tag" | default .LinkTitle }}</title>
+{{ end }}
```
carlmjohnson, line 3: This remove "• Spotlight PA" from the page title. Why override this block? Default should be fine.

layouts/_default/newsletters-signup.html
```diff
@@ -1,3 +1,10 @@
+{{ define "title" }}
+  <title>{{ .Param "title-tag" | default .LinkTitle }}</title>
+{{ end }}
+{{ define "body-class" }}
+  text-[16px] bg-beige font-serif leading-normal text-s-9
+{{ end }}
```
carlmjohnson, line 6: I think if you set the default body-class in baseof, you won't need to add an override here.

carlmjohnson, line 6: I think if you set the default body-class in baseof, you won't need to add an override here.

layouts/partials/tw/nav.html
carlmjohnson: Alpine has two ways to do `:class` bindings. One is a simple string, like you have here, but the other is an object with string keys and bool values. If you do the object way, Alpine will correctly turn off a class when you don't want it anymore. It would look like 

```html
:class="{
  'sm:py-0': toggle,
  'sm:h-11': toggle,
  'md:justify-start': toggle,
  'md:justify-end': ! toggle,
}"
```

layouts/partials/tw/nav.html
carlmjohnson: I think you can just drop the line then, no?

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,311 @@
-<nav
+<div
   id="top"
   x-data="{ isOpen: false }"
```
carlmjohnson, line 3: isOpen is obsolete now that it's replaced with toggle, so you can delete this line.

layouts/partials/tw/nav.html
```diff
@@ -1,137 +1,311 @@
...
-            class="site-nav:svg-icon"
-            viewBox="0 0 20 20"
-            aria-hidden="true"
-            style="top: 1px; margin-right: 2px"
+        {{ if .HasChildren }}
+          <ul
+            class="
+              my-3 flex w-full flex-wrap gap-2 md:h-0 md:w-0 md:overflow-hidden
+            "
+            :class="toggle ? 'md:visible md:overflow-visible md:w-auto md:h-auto sm:w-48' : 'md:hidden md:w-0'"
```
carlmjohnson, line 194: It's probably worth using the `:class` object syntax here too, because if md:visible ever stops overriding md:hidden for some reason, this would break.

carlmjohnson (changes requested):
It looks good on my iPhone SE and my testing Android, so I think it's pretty much ready to go once the body class stuff is fixed.

layouts/partials/tw/nav.html
carlmjohnson: Here are the Alpine docs: https://alpinejs.dev/directives/bind#class-object-syntax

> This technique offers a unique advantage to other methods. When using object-syntax, Alpine will NOT preserve original classes applied to an element's class attribute.
> 
> For example, if you wanted to apply the "hidden" class to an element before Alpine loads, AND use Alpine to toggle its existence you can only achieve that behavior using object-syntax.

Basically by default, Alpine appends new classes to the DOM but leaves the initial ones from the HTML still there, so the only way to turn a class that is in the HTML off is with the object syntax.

layouts/_default/newsletters-signup.html
jeffrummel: body class doesn't transfer over from baseof

layouts/partials/tw/nav.html
jeffrummel: I don't follow. I'm already using the `:class` syntax here, aren't I?

layouts/partials/tw/nav.html
jeffrummel: This is removed: https://github.com/spotlightpa/poor-richard/pull/422/commits/cb0785ca062d3631356995b5fef37ca2b280ff9d

layouts/partials/tw/nav.html
jeffrummel: never mind. I'm getting your comments out of order

carlmjohnson (approved):
LGTM. 🎉 🎉 🚢

PR #429, opened 2023-06-07
Nav touchups

carlmjohnson:
This makes the nav more accessible and fixes issues with tabbing through the invisible items.

### DIFF
--- a/layouts/partials/tw/nav.html
+++ b/layouts/partials/tw/nav.html
@@ -57,15 +57,15 @@
 </div>
 
 <nav
-  x-data="{ toggle: false }"
+  x-data="{ isOpen: false }"
   x-id="['dropdown-button']"
   class="
     relative flex h-11 w-full items-stretch justify-between border-b border-t bg-white px-3 font-sans md:justify-center
   "
-  :class="toggle ? 'px-5 sm:px-3' : ''"
-  x-on:keydown.escape.prevent.stop="toggle = false; $refs.button.focus()"
-  @click.outside="toggle = false; $refs.button()"
-  @focusin.outside="toggle = false"
+  :class="isOpen ? 'px-5 sm:px-3' : ''"
+  x-on:keydown.escape.prevent.stop="isOpen = false; $refs.button.focus()"
+  @click.outside="isOpen = false; $refs.button()"
+  @focusin.outside="isOpen = false"
 >
   <a
     href="/search/"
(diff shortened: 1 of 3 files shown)


layouts/partials/tw/nav.html
```diff
@@ -1,25 +1,33 @@
...
     >
       <svg class="block h-8 w-full sm:h-14 md:h-16 lg:h-20">
-        <title>Home</title>
-        <desc>Spotlight PA logo</desc>
+        <title>Spotlight PA Home</title>
         {{ partial "helper/svg.html" "banner" }}
       </svg>
     </a>
-  </h1>
+  </div>
```
jeffrummel, line 30: What's the benefit of using a div over h1 here? Using an H1 to define the title in the page header has historically been a best-practice

carlmjohnson: We use H1 for the page title already, so now we have two H1s, one that says Spotlight PA and one that says whatever the page is about.

layouts/partials/tw/nav.html
```diff
@@ -1,25 +1,33 @@
...
 >
-  <h1
+  <div
+    class="hidden text-xs font-extrabold leading-tight tracking-wide sm:block sm:w-32 lg:w-52 lg:text-sm xl:text-base"
+  >
+    <p>Independent</p>
+    <p>Nonpartisan</p>
+    <p>Nonprofit</p>
+  </div>
+  <div
```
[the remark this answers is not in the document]
carlmjohnson: I just did another test, and I think you're right. It's odd to hear the secondary tagline without context. I'll switch it back.

carlmjohnson:
Margins should be fixed once it deploys.

jeffrummel:
Screenshot 2023-06-08 at 9 16 00 AM


menu no longer aligns with max page width

jeffrummel:
Screenshot 2023-06-08 at 09-34-31 Spotlight PA Pennsylvania news and investigations

There's also a positioning problem with expanded menu on larger screens

carlmjohnson:
The company needs to buy me a larger laptop. Mine doesn't get that wide. 😆

carlmjohnson:
Okay, I think that fixes it. Definitely feeling the CSS meme right now.

carlmjohnson:
Seems okay with one more:

image

This what I get if I just add an element in the DOM inspector:

image

I can turn off the transition-translation if it's too distracting.

PR #435, opened 2023-06-30
Use SVG partial for copy button

carlmjohnson:
I also did some minor upgrades. 

I found out we can't move to the current Tailwind because they added in CSS logical properties with thing-s-1 for thing-start: .25rem, which conflicts with our s ("silver") colors. We'll have to rename it to g ("gray") before we do an upgrade.


carlmjohnson:
https://deploy-preview-435--poor-richard-spotlightpa.netlify.app/embeds/

jeffrummel (approved):

PR #438, opened 2023-07-07
Miscellaneous changes

carlmjohnson:
- Deletes bulma.css
- Fixes rounding on search input box
- Changes headline sizes for HP featured item to try to fit smaller screens better.

Let me know if you think the size changes work. I think it should be better on tablet-y sizes.


carlmjohnson:
Now: 
image

Updated:

image

PR #442, opened 2023-07-12
meta.html: Add Twitter creator and Slack metadata

carlmjohnson:
H/t: See https://www.trysmudford.com/blog/fancy-slack-meta-tags/


carlmjohnson:
image

Works in Slack.

PR #450, opened 2023-08-17
Jr/home takeover

jeffrummel:


layouts/_default/baseof.html
```diff
@@ -89,7 +89,13 @@
     {{ block "modal" . }}
       {{ if .Param "modal-exclude" | not }}
         {{ if .Param "sticky-active" }}
-          {{ partial "tw/modal-sticky.html" . }}
+          {{ partial "tw/modal-takeover.html" ( dict
+            "filename" "2023/08/01jd-w2vj-77eb-3dp1.jpeg"
+            "hed" "Time is running out and we need your help"
+            "dek" "Only a few hours remain for your gift to Spotlight PA to be DOUBLED. Contribute now and ensure our trusted, nonpartisan investigative journalism can continue."
+            "link" "https://spotlightpa.fundjournalism.org/statecollege/?campaign=701Dn000000YqKfIAK"
```
carlmjohnson, line 96: Is it supposed to be a statecollege donate link?

layouts/partials/tw/modal-takeover.html
```diff
@@ -0,0 +1,121 @@
...
+      >
+        {{ $hed }}
+      </h2>
+    </header>
+    <span class="c-red-8 -order-1 self-end sm:order-1 sm:self-start md:mt-28">
+      <svg
+        class="fill-red-8 sm:origin-center sm:scale-[(-1,_1)]"
+        width="16"
+        height="22"
+      >
```
carlmjohnson, line 58: Maybe add a comment like `<!-- triangle -->` so it's clear that it's the little fold out triangle.

config/_default/params.json
```diff
@@ -20,7 +20,8 @@
   "footer-promo-mobile-height": 250,
   "footer-promo-mobile-images": [],
   "footer-promo-mobile-width": 500,
-  "newsletter-active": true,
+  "newsletter-active": false,
+  "takeover-active": true,
```
carlmjohnson, line 24: I’m going to have to add a slot to Almanack to keep this from being overwritten. Seems like a good idea though so Chris can toggle it.

src/utils/metrics.js
```diff
@@ -14,7 +14,7 @@ const SAW_DONATE_MODAL_KEY = "saw-donate-modal-totebag";
 const SIGNED_UP_FOR_NEWSLETTER_KEY = "signed-up-for-newsletter";
 const PRIOR_FUNNEL_STATUS_KEY = "funnel-status";
 
-const SHOW_INTERVAL = 7 * 24 * 60 * 60 * 1000; // 1 week
+const SHOW_INTERVAL = 10 * 60 * 1000; // 1 minutes
```
carlmjohnson, line 17: Should say 10 minutes.

layouts/partials/tw/modal-takeover.html
```diff
@@ -0,0 +1,132 @@
+{{ $imagename := .filename }}
+{{ $hed := .hed }}
+{{ $dek := .dek }}
+{{ $link := .link }}
+
+
+<div x-data x-intersect.once="$dispatch('open-takeover')"></div>
```
carlmjohnson, line 7: If it’s being trigger immediately, you can remove this and add an x-intersect.once="show()" to the section and remove the @ listener.

layouts/_default/baseof.html
```diff
@@ -41,6 +41,15 @@
     data-page-cat="{{ $pageCat }}"
     {{ block `body-attributes` . }}{{ end }}
   >
+    {{ if .Param "takeover-active" }}
+      {{ partial "tw/modal-takeover.html" ( dict
+        "filename" "2023/08/01jd-w2vj-77eb-3dp1.jpeg"
+        "hed" "Time is running out and we need your help"
+        "dek" "Only a few hours remain for your gift to Spotlight PA to be DOUBLED. Contribute now and ensure our trusted, nonpartisan investigative journalism can continue."
+        "link" "https://spotlightpa.fundjournalism.org/statecollege/?campaign=701Dn000000YqKfIAK"
```
carlmjohnson, line 49: Did you see my comment before asking if this is the right donate link?

src/utils/metrics.js
carlmjohnson: Probably we should have a separate thing for the takeover because I think this will hide from people it thinks are already subscribers.

layouts/partials/tw/modal-takeover.html
jeffrummel: Removing line 7 and the @ listener, then adding in `x-intersect.once="show()"` isn't working.

layouts/_default/baseof.html
jeffrummel: This is the link provided to me.

src/utils/metrics.js
```diff
@@ -13,8 +13,10 @@ const FROM_MC_KEY = "originated-from-mailchimp";
 const SAW_DONATE_MODAL_KEY = "saw-donate-modal-totebag";
```
carlmjohnson, line 13: Wow, that key is really old. I forgot about the totebag promo. 😆

src/utils/metrics.js
```diff
@@ -13,8 +13,10 @@ const FROM_MC_KEY = "originated-from-mailchimp";
...
 const SIGNED_UP_FOR_NEWSLETTER_KEY = "signed-up-for-newsletter";
 const PRIOR_FUNNEL_STATUS_KEY = "funnel-status";
+const SAW_TAKEOVER_MODAL_KEY = "saw-newsletter-modal";
```
carlmjohnson, line 16: That should be `saw-takeover-modal`.

src/utils/metrics.js
```diff
@@ -75,6 +77,19 @@ let shouldShowModalNewsletter = (() => {
   return now - lastPrompt > SHOW_INTERVAL;
 })();
 
+let shouldShowModalTakeover = (() => {
+  if (onTestNewsletterPage) {
+    return true;
+  }
+  let cameFromMCOn = loadDate(FROM_MC_KEY) || 0;
```
carlmjohnson, line 84: You can delete these lives from shouldShowModalTakeover because we don't care if you came from Mailchimp, we still want you to see the takeover.

src/utils/metrics.js
```diff
@@ -75,6 +77,19 @@ let shouldShowModalNewsletter = (() => {
...
+  let sawNLModalOn = loadDate(SAW_NEWSLETTER_MODAL_KEY) || 0;
+  if (!cameFromMCOn && !sawNLModalOn) {
+    return true;
+  }
+  let lastPrompt = Math.max(cameFromMCOn, sawNLModalOn);
```
carlmjohnson, line 89: I think it should be this (not tested):

```js
let shouldShowModalTakeover = (() => {
  if (onTestNewsletterPage) {
    return true;
  }
  let sawTakeoverModalOn = loadDate(SAW_TAKEOVER_MODAL_KEY) || 0;
  return now - sawTakeoverModalOn > TAKEOVER_INTERVAL;
})();
```

src/utils/metrics.js
```diff
@@ -84,10 +99,22 @@ export let modalKind = (() => {
...
 })();
 
+export let modalTakeover = (() => {
+  if (window.matchMedia("(prefers-reduced-motion), (speech)").matches) {
+    // eslint-disable-next-line no-console
+    console.warn("prefers-reduced-motion; aborting modal display");
+    return "none";
+  }
+  return shouldShowModalTakeover ? "takeover" : "sticky";
+})();
```
carlmjohnson, line 109: Instead of having a new variable for modalTakeover, I think this should incorporated into modalKind, like this:

```js
export let modalKind = (() => {
  if (window.matchMedia("(prefers-reduced-motion), (speech)").matches) {
    // eslint-disable-next-line no-console
    console.warn("prefers-reduced-motion; aborting modal display");
    return "none";
  }
  if (shouldShowModalTakeover) {
    return "takeover";
  }
  if (shouldShowModalNewsletter) {
    return "newsletter";
  }
  return "sticky";
})();
```

carlmjohnson: Hmm, that won't work because the JS doesn't know if the takeover is turned on or off. Leave it your way for now.

src/utils/metrics.js
jeffrummel: This works 👍

src/utils/metrics.js
jeffrummel: Updated: https://github.com/spotlightpa/poor-richard/pull/450/commits/17bfd2ecccea2f8fc4945d53975b6a5659be3b0c

PR #453, opened 2023-08-18
Promos: Check modal is on page before saying it should be triggered

carlmjohnson:
- Let Almanack overwrite properties of takeover
- Fix a bug with recording subscriber status when clicking "already a member"


jeffrummel (approved):
Looks good!

PR #454, opened 2023-08-21
Impact Page

jeffrummel:


layouts/partials/tw/impact-carousel-panels.html
```diff
@@ -0,0 +1,117 @@
+<!-- Panels -->
+<div
+  class="relative col-span-8 col-start-1 row-span-3 row-start-1 w-full items-center justify-center md:col-span-14"
+>
+  <!-- PANEL NAVIGATION -->
+  <button
+    class="absolute left-0 top-0 z-20 h-full w-16 text-6xl text-white"
+    @click="select($el.id) - 1"
```
carlmjohnson, line 8: Should that be `select($el.id - 1)`? I think select() doesn't return anything, so it would just be the same as `undefined - 1`. 

 I'm not sure `select($el.id - 1)` will work either because id is a string and not a number. Worth a shot though. 

In the featured/slider, I have currentSlide be a number and then there is `this.$watch("currentSlide"` call that tells Alpine to scroll over to whatever currentSlide gets set to whenever it changes.

That one cheats a bit by making the arrow buttons not actually do anything. Instead when you click an arrow, it falls through to the underlying slide and then that slide gets the click and brings itself into view. But that aside, I think having the currentPanel be a number makes some of the other stuff easier.

carlmjohnson:
image

Can't see the right arrow here because it's white on white. Maybe add a shadow or a ramp?

layouts/partials/blocks/social.html
```diff
@@ -0,0 +1,49 @@
+{{ with .block }}
+  <nav
+    class="mx-5 mt-10 rounded border border-g-4 p-6 shadow md:mx-auto md:mt-16 md:max-w-screen-lg md:px-20 md:py-10"
+    data-ga-category="follow"
+  >
+    <h3
+      class="font-sans text-base font-bold uppercase leading-none tracking-wider text-orange"
```
carlmjohnson, line 7: We don't really use text-orange many places anymore. Maybe this should be a different color?

carlmjohnson (commented):
Looks really good.

carlmjohnson (commented):
image

On mobile, it's getting put into a weird aspect ratio because the text doesn't fit. I'm not sure what the best solution is here. Maybe just tell the background image to fill the space?

PR #465, opened 2023-09-01
Impact page JS fixes

carlmjohnson:


carlmjohnson:
I started by trying to use the Alpine.js Tab component, but it didn't quite fit what we were doing, so I ended up just writing a short thing that has the current panel number. It's not perfect because you can't have two of these on the page without stomping on each other, but it works for now.

jeffrummel (approved):
LGTM

PR #467, opened 2023-09-06
Jr/election 2023

jeffrummel:


carlmjohnson:
We should probably add a footer link to elections.

carlmjohnson:
When you jump to an item, the top of it gets cut off by the sticky nav.

Screenshot 2023-09-06 at 9 00 18 AM

To workaround this problem, you canadd a [scroll-margin](https://tailwindcss.com/docs/scroll-margin) to each target that is the size of the sticky nav. Here's the same thing, but with a `scroll-mt-[82px]` class on the #events div:

image

It seems like there should be a way to just add one class and tell it that every thing needs a scroll margin because of the sticky nav, but I haven't found a way to do that. Maybe something with scroll-padding?

carlmjohnson:
This guy says scroll-padding has to go directly on the `<html>` because that's the scroll container: https://markus.oberlehner.net/blog/simple-solution-for-anchor-links-behind-sticky-headers/. I guess that makes sense. It worked when I changed the "html-class" block of layouts/_default/elections.html to have `scroll-pt-[82px]`.

config/_default/menus.toml
```diff
@@ -206,7 +206,12 @@ weight = 10
 [[news]]
 name = "Events"
 url = "/events/"
-weight = 11
+weight = 12
+
+[[news]]
+name = "Elections"
+url = "/election/"
+weight = 8
```
carlmjohnson, line 214: The Capitol is also 8, so one of them should be 9.

layouts/partials/tw/internal-page-nav.html
```diff
@@ -0,0 +1,60 @@
+<div
+  class="mx-auto flex max-w-screen-xl  flex-col md:flex-row"
+  data-ga-category="sticky"
+>
+  <div class="no-js:hidden relative mx-auto w-96 max-w-full md:hidden">
+    <select
+      x-data
+      @change="
+        window.location.assign($event.target.value);
+        $report({
```
carlmjohnson, line 10: I think $report is broken with the current analytics setup. Just remove this part and leave the `window.location.assign($event.target.value);` part.

PR #469, opened 2023-09-06
adds chris to board page at position one

jeffrummel:


carlmjohnson:
If Chris is number one on the staff page, we don't need to sink Heather Adams to last anymore.

PR #470, opened 2023-09-06
adds title to Annie Madonia

jeffrummel:


carlmjohnson:
Hugo has a concept of weight for sorting, but we're not using it since we sort by lastName, but if we're getting multiple exceptions, maybe we should see if we can sort by lastName then by weight as an override.

jeffrummel:
> Hugo has a concept of weight for sorting, but we're not using it since we sort by lastName, but if we're getting multiple exceptions, maybe we should see if we can sort by lastName then by weight as an override.

Good thinking. I can circle back to this when I dig into redesigning the board page.

PR #471, opened 2023-09-08
requested updates:

jeffrummel:
- sets fixed height with view more toggle on bios
- adds title to Michael Berry

### DIFF
--- a/layouts/_default/board-of-directors-page.html
+++ b/layouts/_default/board-of-directors-page.html
@@ -52,7 +52,7 @@
   <section class="bg-s-2 px-2 pb-10  font-sans">
     <div class="mx-auto my-14 w-full max-w-screen-xl">
       <ul
-        class="mx-auto my-14 mt-10 grid w-full grid-flow-dense gap-8 md:grid-cols-2"
+        class="mx-auto my-14 mt-10 flex w-full flex-wrap gap-8"
         data-ga-category="inline-footer:author"
       >
         {{ range $board }}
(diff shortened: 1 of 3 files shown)


layouts/_default/board-of-directors-page.html
```diff
@@ -67,8 +67,49 @@ <h2 class="mb-5 text-5xl font-black">{{ . }}</h2>
...
+            {{ partial "helper/svg.html" "plus" }}
+          </svg>
+        </span>
+        <span class="ml-2 align-bottom uppercase hover:underline">
+          Show more…
+        </span>
+      </button>
+      <button
+        x-show="isOpen"
+        :aria-expanded="''+isOpen"
```
jeffrummel, line 97: @carlmjohnson This button closes the bio to show less. What's the correct aria label here?

carlmjohnson:
image

Her bio is too short to need show more, but it has the button anyway. Maybe just make the boxes smaller so that she needs a show more button too? Every once in a while we'll have a feature with a very short sidebar that has this issue too, but it's not so common.

layouts/_default/board-of-directors-page.html
carlmjohnson: You could add an ID to the "Show less…" span and then do `aria-label="{{ $showLessID }}"` here.

carlmjohnson: Does it need an aria-label? It might just be okay without it, since that's the content anyway…

jeffrummel: These aria labels are removed

PR #476, opened 2023-09-13
Jr/recaptcha

jeffrummel:


carlmjohnson (approved):
We should make a partial to reduce the duplication like

```html
<input
  type="hidden"
  name="redirect"
  value="https://www.spotlightpa.org/newsletters/thanks/"
/>
<input
  type="hidden"
   name="redirect_sorry"
   value="https://www.spotlightpa.org/newsletters/sorry/"
 />
 <input type="hidden" x-data :name="'shibboleth'" :value="'PA Rocks!'" />
 <input type="hidden" name="recaptcha_response" class="recaptcha-response" />
 ```

But let's just do this first. 

I'll need to change the email alerts server to verify the captcha value and then we can drop the shibboleth field.

What happens if the captcha thinks you're a bot? Does it ever pop up a box or just silently disapprove of you?

jeffrummel:
> What happens if the captcha thinks you're a bot? Does it ever pop up a box or just silently disapprove of you?

According to the docs we never get a box, but the validity of the submission is logged in Captcha and fails silently

carlmjohnson:
Also, the little recaptcha floater is showing up everywhere. We should try not to let it pop up until we need it.

image

carlmjohnson:
I've been trying to rewrite the JS to only trigger when needed so the token doesn't expire, but I've been hitting different road blocks. Let's wait for the meeting and see if they still want to do the recaptcha now that the honeypot is blocking so much stuff already.

PR #478, opened 2023-09-14
Newsletters: Fix berks county

carlmjohnson:


carlmjohnson:
I'm going to do this now and follow up with a fix to make a master set of hidden fields.

PR #484, opened 2023-09-28
CSS upgrades

carlmjohnson:
I overhauled the CSS to work with new versions of Tailwind. This involved renaming the s color to g because new Tailwind uses -s- for start positioning. Now when you add a new class to the page that is not used elsewhere on the site, Hugo should reload twice: once to add the HTML and again to add the new class to Tailwind. Seems to work. It also means now we can load the page CSS without JavaScript in dev.


jeffrummel:
The breadcrumbs "succeeds" arrow doesn't render in the breadcrumbs: `before:content-['\0227B']`

I'm not sure if this has to do with the upgrade or with Raleway. Its absence an be seen here: https://6516cb839eaea90008a6d9c8--poor-richard-spotlightpa.netlify.app/about/mission/

carlmjohnson:
I actually noticed there was no chevron between breadcrumbs, but I didn't realize it was a regression. I just thought that was the current design. I'll work on a fix.

PR #488, opened 2023-10-04
Fix smooth scroll

carlmjohnson:


layouts/shortcodes/custom/courts-list.html
```diff
@@ -11,10 +11,22 @@
...
+    @click.prevent="
+      selected = issue;
+      window.setTimeout(() => {
+        document.getElementById(issue).
+          scrollIntoView({
+          behavior: 'smooth',
+          block: 'start',
+          inline: 'start'
+        });
+      }, 250);
```
carlmjohnson, line 27: It waits 250ms, then it smooth scrolls the ID into view.

jeffrummel (approved):

PR #494, opened 2023-10-12
JS: Get scroll to play working

carlmjohnson:
This seems like it's working now, at least on desktop. I'm a little worried about if it's going to work on mobile, because it might be more strict there about requiring a touch event to trigger play.


carlmjohnson:
I notice that when I scroll out, the text disappears, and then the page jerks up. Maybe to fix that, we should change it to not trigger `onScreen = false` on x-intersect:leave.

jeffrummel (approved):

PR #497, opened 2023-10-18
sets z-index on nav

jeffrummel:


carlmjohnson:
I changed the Z index when I was upgrading Tailwind because I saw some overlaps in places. It's tricky to get right because there are a lot things we either want to be under or over depending on the page.

carlmjohnson:
If you open the menu and then click on search, the menu is too high now.

image

Also, I think moving the nav up means the search icon is over the modal. 

We may just need to bite the bullet and move a lot of zindexes, not just the nav.

carlmjohnson:
image

Newsletter modal too.

jeffrummel:
@carlmjohnson Let me know if I'm good to merge here https://github.com/spotlightpa/poor-richard/pull/497/commits/4474ebf5c2f75a8e4fc03d5323483aa18f87e580

carlmjohnson:
Looks good. I think we can ship this.

I can still get into a weird situation if I got to the about page, open the menu, scroll down the sticky nav, then scroll up again, but I'm not sure it's worth fixing because that's a weird thing to do.

image

It's a pretty minor visual bug all things considered. If we do fix it, it should probably be by just telling the menu to close when the sticky nav opens instead of finding another z-slot for things.

layouts/shortcodes/featured/main-takeover.html
```diff
@@ -10,7 +10,7 @@
 {{ $hiddenField := .Get "hiddenField" }}
 {{ $story := .Get "story" }}
 <header
-  class="relative z-20 flex min-h-[80vh] flex-col justify-center bg-black text-white"
+  class="relativez-50 flex min-h-[80vh] flex-col justify-center bg-black text-white"
```
carlmjohnson, line 13: This is a typo.

PR #504, opened 2023-10-26
Elections update for expired events

jeffrummel:


layouts/shortcodes/sidebar-link.html
```diff
@@ -37,6 +38,18 @@
         >Add to Calendar</span
       >
     </button>
+  {{ else if $dateExpired }}
```
carlmjohnson, line 41: Can't we just test if today is after the date?

carlmjohnson: `time "2023-08-01"` turns a date string into time object. Then you can compare it with the `now` time object. Something like

```
{{ $dateObj := time $date }}
{{ if now.After $dateObj }}
 ```

https://pkg.go.dev/time#Time.After documents the methods on a time object.

jeffrummel: hugo can't seem to parse the date from the shortcode. 

`{{ $date := "2023-08-01" }}` works, but `{{ $date := .Get "date" }}` does not.

carlmjohnson: I did a quick test and I ran into the problem that the build fails if `time $date` is "", but I think it should be okay if it's behind an `if $date` statement.

carlmjohnson: Opened a sub-PR.

carlmjohnson:
I'm going to go ahead and merge it just so it's done before the weekend.

PR #515, opened 2023-11-14
Partners: Fix up JS

carlmjohnson:
I think this should be working now. There was a pre-existing counties.json from some old project that we aren't using anymore, so I added the translate coordinates to it. Also two partners are not in any county, so I moved them to N/A. This looks really cool. One of the dots in the north is over a county that doesn't have any partners listed. Do you know which partner that is supposed to be?


carlmjohnson:
Some of the partners are listed in multiple counties, which is a little confusing.

carlmjohnson:
I think maybe the dots are misaligned a little?

jeffrummel (approved):

PR #517, opened 2023-11-15
updates tagline

jeffrummel:


carlmjohnson (approved):
Neat. I didn't know we were getting a trademark on that.

PR #519, opened 2023-11-17
adds slider for standard stories

jeffrummel:
@carlmjohnson on the last slide all elements inside .article-content gets pushed to the left on lg viewport. removing `overflow-hidden` fixes the issue, but I'm not sure what display(s) that will break on other stories. Feel free to take a look if you have the time.

Screenshot 2023-11-17 at 2 45 12 PM


carlmjohnson:
The content well has a hard limit around it so that it won't have weird horizontal scrolling stuff on mobile when some random thing is the wrong size. There are like 5 "okay overflow fixed for real this time!" commits somewhere in the Git history.

Maybe the solution here is just to put a border around the gallery like this but not as ugly:

image

jeffrummel:
The issue is that the body copy is getting cropped on the left-side when you scroll to the last slide. You can see an example of this here: https://slidertest--poor-richard-spotlightpa.netlify.app/news/2030/12/standard-example-xxx/

I'm getting consistent results on FF and Chrome

carlmjohnson:
Hmm, I didn’t see it on Safari. Maybe it’s because they don’t support all the containment stuff.

earthboundkid:
image

I like the new style, but this broke the old carousels like https://slidertest--poor-richard-spotlightpa.netlify.app/news/2021/06/pa-medical-marijuana-insurance-drug-treatment-confusion/. If you look at the JS console, it's also complaining about not having access to the disableNextAndPreviousButtons JS, I guess because its parent component doesn't have the right Alpine x-data.

jeffrummel:
> image
> 
> I like the new style, but this broke the old carousels like https://slidertest--poor-richard-spotlightpa.netlify.app/news/2021/06/pa-medical-marijuana-insurance-drug-treatment-confusion/. If you look at the JS console, it's also complaining about not having access to the disableNextAndPreviousButtons JS, I guess because its parent component doesn't have the right Alpine x-data.

@earthboundkid Here's an update with the new design on both sliders:

- https://slidertest--poor-richard-spotlightpa.netlify.app/news/2021/06/pa-medical-marijuana-insurance-drug-treatment-confusion/
- https://slidertest--poor-richard-spotlightpa.netlify.app/news/2030/12/standard-example-xxx/

LMK if this works or if you'd prefer I go back to the original design.

layouts/shortcodes/featured/slide.html
```diff
@@ -21,40 +21,38 @@
...
+<li
+  x-bind="disableNextAndPreviousButtons"
+  role="option"
+  class="h-full w-11/12 flex-[0_0_auto] snap-center px-0 font-bold md:w-11/12 [&:first-child]:snap-start [&:last-child]:snap-end"
 >
-  <div class="relative h-0 w-full pb-16x9">
-    <picture
-      class="
+  <figure
+    class="group flex-none cursor-pointer snap-center pb-2 focus:outline-none"
```
earthboundkid, line 30: Can you make it so it only uses cursor-pointer over the left/right select buttons?

layouts/shortcodes/featured/slider.html
```diff
@@ -1,47 +1,100 @@
 {{ $id := partial "helper/new-id" . }}
```
earthboundkid, line 1: slider and featured/slider almost identical except for the container class. Can we have them both call say partials/tw/slider.html and pass in a .containerClass or something like that to reduce the duplication?

earthboundkid (approved):
Very cool!

PR #545, opened 2024-01-09
Related Stories shortcode

jeffrummel:
**Sample Shortcode:**
```
{{<featured/related-stories 
  title="More Stories from the 2024 Election" 
  link="/election"
  cta="Visit the Election Center →"
  limit="12"
>}}
```


layouts/shortcodes/featured/related-stories.html
```diff
@@ -0,0 +1,200 @@
...
+{{ end }}
+{{ $pages := $topic.Pages.ByPublishDate }}
+{{ if gt $pages.Len 1 }}
+  {{ $jsOpts := dict
+    "minify" true
+  }}
+  {{ $js := resources.Get "js/esbuild/smoothscroll.js" | js.Build $jsOpts }}
+  <script type="module">
+    {{ $js.Content | safeJS }}
+  </script>
```
carlmjohnson, line 31: If there’s more than one related block on the page, it will execute this twice, but if we change it to be src=$js.RelPermalink and keep type=module, it should only execute once even if it is included multiple times.

carlmjohnson (approved):

carlmjohnson:
Idea for the backlog: We could make a widget at /embed/ to build related article shortcodes.

PR #552, opened 2024-01-19
Feeds: Add elections.json feed for Dewey

carlmjohnson:


jeffrummel (approved):

ScotterC:
🙏  Thanks for the JSON feed!

@earthboundkid the ID in the feed is the permalink for the post - is that actually permanent? Is that the internal ID used as well? I imagine URL slugs can change on title update or manual edits. Dewey will be looking for updates to content over time and will need a primary key of sorts to find the latest content in the feed.

Also, a nice to have would be being able to retrieve each post by it's ID - /feeds/{id}.json or xml. It's not critical but means not having to load the entire data feed to check for updates to each post.

PR #554, opened 2024-01-22
adds election contest page

jeffrummel:


prettier-html-ignore.rc
```diff
@@ -3,7 +3,8 @@ layouts/partials/featured/content-area-start.html
 layouts/partials/header/meta.html
 layouts/partials/tw/image-rotator.html
 layouts/partials/tw/promo-double.html
-layouts/partials/tw/topper.html
+layouts/partials/tw/promo-double.html
```
carlmjohnson, line 6: This is a duplicate line.

carlmjohnson (changes requested):
The error message from prettier seems to mean there's something unbalanced in the HTML. I'd like to track it down if we can. I found some spots with `/ style=aspect-ratio:128/72` but fixing that didn't balance the HTML.

jeffrummel:
> The error message from prettier seems to mean there's something unbalanced in the HTML. I'd like to track it down if we can. I found some spots with / style=aspect-ratio:128/72 but fixing that didn't balance the HTML.

I believe this is caused by missing parentheses that are stripped at build time. `style=aspect-ratio:128/72` doesn't pass validation in prettier but `style="aspect-ratio:128/72"` does. I spent some time today adding parentheses back into the links hoping that would be enough but it still wouldn't pass validation and the parentheses were stripped out on the deploy.

content/elections-contest.md
```diff
@@ -0,0 +1,197 @@
...
+link="https://www.spotlightpa.org/news/2023/10/pennsylvania-election-2023-voting-rights-event/"
+
+[[blocks.entries]]
+hed="Result Review"
+dek="And the winners are … Join us to learn about how the judicial election results will influence politics and policy in the years to come. Registration coming soon."
+date="2023-11-16"
+dateVerbose="Nov. 16, 2023 on Zoom"
+link="https://www.spotlightpa.org/news/2023/10/pennsylvania-election-2023-results-event/"
+registration="https://zoom.us/webinar/register/WN_wf_P08sRQF-jRFrnG-T4NA"
++++
```
carlmjohnson, line 197: The blocks in here don't really work, do they? Maybe just drop them so we don't try to tweak it and forget why it's not working in 6 months.

carlmjohnson (approved):

PR #569, opened 2024-02-15
list.html: Fix river logic

earthboundkid:


layouts/_default/list.html
```diff
@@ -180,14 +180,9 @@ <h2 class="text-2xl font-extrabold">
       data-ga-category="river"
       data-sf-campaign="701Dn000000YgodIAC"
     >
-      <nav
-        x-data="readmore(true)"
-        x-ref="root"
-        @mouseover.once="load()"
-        @touchstart.passive.once="load()"
-      >
+      <nav>
```
earthboundkid, line 183: This page doesn't have a readmore button, so this was throwing an error.

layouts/_default/list.html
```diff
@@ -159,6 +155,10 @@ <h2 class="text-2xl font-extrabold">
 {{ define "extra-main" }}
   {{ $p := partial "helper/paginator" . }}
   {{ $pages := $p.Pages }}
+  {{ $pinnedPages := where $pages "Params.pinned" true }}
+  {{ $featuredPages := union $pinnedPages $pages | first 5 }}
+  {{ $river := complement $featuredPages $pages }}
```
earthboundkid, line 160: Using `complement` here fixes the case of items from further down the river jumping to the top after pinning.

PR #579, opened 2024-02-20
Jr/fix search z index

jeffrummel:


jeffrummel:
@earthboundkid These custom z-indexes are not ideal and I've added a to-do for myself to go back and refactor all uses of z-index for down the road.

earthboundkid (approved):
Yeah, maybe we should make a file like docs/z-index.md or something and try to map out all the z-indexes we need there, so we have one place to consult as new things are added or removed.

PR #584, opened 2024-02-26
Cleanup dead templates

earthboundkid:


earthboundkid:
LMK if you think any of these are still in use. I think they're all dead. There's a --printUnusedTemplates option to hugo but it gets false positives from the promos we aren't using.

jeffrummel (approved):
I agree and think these are all unnecessary.

PR #589, opened 2024-02-28
adds complement to topic stories

jeffrummel:


layouts/partials/tw/list-callout-story.html
```diff
@@ -3,7 +3,9 @@
 {{ if gt $rows 0 }}
   {{ $limit = mul $rows 1 }}
 {{ end }}
+{{ $featuredStory := partial "helper/get-picks" "featuredStories" }}
 {{ $page := .pages }}
+{{ $pages := index $page }}
```
earthboundkid, line 8: Another one here.

layouts/partials/tw/list-callout-story.html
```diff
@@ -22,7 +24,7 @@
     </h2>
   {{ end }}
 
-  {{ range $page | first 1 }}
+  {{ range $pages | first 1 | complement $featuredStory }}
```
earthboundkid, line 27: If it's in this order, it might end up empty. I would do `$pages | complement $featuredStory | after 1`.

earthboundkid (approved):

PR #594, opened 2024-03-01
removes if limit statement, adds complement for pinned stories

jeffrummel:


layouts/partials/tw/list-callout-story.html
```diff
@@ -1,13 +1,6 @@
-{{ $rows := .rows | default 2 }}
-{{ $limit := 0 }}
-{{ if gt $rows 0 }}
-  {{ $limit = mul $rows 1 }}
-{{ end }}
```
jeffrummel, line 5: @earthboundkid as best I can tell, the `$rows` and `$limit` weren't doing anything but breaking our complements.

earthboundkid (approved):

PR #599, opened 2024-03-05
events.html: Fix thumb for upcoming events; complement river

earthboundkid:


layouts/_default/events.html
```diff
@@ -9,13 +9,12 @@
   {{ $p := partial "helper/paginator" . }}
   {{ $pages := $p.Pages }}
   {{ $upcomingEvents := where $pages `Params.event-date` "gt" now }}
-  {{ $upcomingEvents = sort $upcomingEvents `Params.event-date` "desc" }}
+  {{ $upcomingEvents = sort $upcomingEvents `Params.event-date` "asc" }}
```
earthboundkid, line 12: I think we want the soonest event at the top of the page.

layouts/_default/events.html
```diff
@@ -51,68 +68,90 @@ <h2 class="text-2xl font-extrabold">
...
-    {{ range after 1 . }}
-      {{ $params := partialCached "helper/page-params" . .RelPermalink }}
-      {{ $subitems = $subitems | append (dict
-        "link" $params.link
-        "title" .Title
-        )
-      }}
-    {{ end }}
-    {{ $imgOpts := dict
-      "filename" $params.image
```
earthboundkid, line 83: This was a bug because the $params object doesn't have image on it. It's on the page.Params.

jeffrummel: I'm not sure I understand the different use-cases for `$params` vs just `$event`

jeffrummel (approved):

layouts/_default/events.html
earthboundkid: $params is the `partialCached "helper/page-params"` object that does some logic for us of calculating the pretty event date, author name, eyebrow links, etc. It exists so we have one place that does the fallback logic for that stuff when a parameter is missing or needs to fallback to different stuff. $event.Params is the native Hugo params dictionary.

PR #604, opened 2024-03-18
AG Quiz

jeffrummel:


content/quiz-results.md
```diff
@@ -0,0 +1,29 @@
...
+link-title = "AG Quiz Results"
+title-tag = "AG Quiz Results"
+description = "AG Quiz Results"
+image = "2024/03/01k0-8hfj-zvrj-6vd3.jpeg"
+image-description = "AG Quiz Results"
+image-credit = ""
+layout = "page-landing"
+signupHed = "Join our mailing list "
+signupDek = "Sign up to receive a weekly review of articles from Spotlight PA and occasional updates about our planning project in Berks County."
+url = "/elections-2024/candidate-quiz/results"
```
earthboundkid, line 12: Add `no-index = true` here.

layouts/partials/blocks/candidate-results.html
```diff
@@ -0,0 +1,80 @@
...
+        title: obj.title
+      }));
+
+      let candidates = [];
+      for (let i= 0; i<orderedCandidates.length; i++) {
+          if (orderedCandidates[i].points >= 0 ) {
+              candidates = [...candidates, orderedCandidates[i]];
+          }
+      }
+      candidates.sort((b, a) => parseFloat(a.points) - parseFloat(b.points));
```
earthboundkid, line 34: Do you need parseFloat if it's already a Number?

layouts/partials/blocks/candidate-results.html
```diff
@@ -0,0 +1,80 @@
+{{ with .block }}
+  {{ $data := index site.Data.candidates2024 }}
+
+  <script>
+    function candidates() {
+
+      let urlParams = new URL(window.location).searchParams.entries();
+      let candidateData = {{$data}}
+
+      function paramsToObject() {
```
earthboundkid, line 10: I think this is the same as [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries), so you could write this instead:

```js
let candidateScores = Object.fromEntries(new URL(window.location).searchParams.entries())
```

Test it out and LMK if it works.

layouts/partials/blocks/candidate-results.html
```diff
@@ -0,0 +1,80 @@
+{{ with .block }}
+  {{ $data := index site.Data.candidates2024 }}
+
+  <script>
+    function candidates() {
+
+      let urlParams = new URL(window.location).searchParams.entries();
+      let candidateData = {{$data}}
```
earthboundkid, line 8: This is probably I don't like mixing JS and template data because it gets hairy fast, so I would probably do something like 

```html
<script data-candidates="{{ $data }}">
    function candidates() {
      let candidateData = JSON.parse(document.currentScript.dataset.candidates);
```

Probably not worth changing though because it's a one off.

layouts/partials/blocks/candidate-results.html
```diff
@@ -0,0 +1,80 @@
...
+
+      let orderedCandidates = candidateData.map(obj => ({
+        name: obj.name,
+        id: obj.id,
+        points: Number(candidateScores[obj.id]),
+        image: obj.image,
+        title: obj.title
+      }));
+
+      let candidates = [];
```
earthboundkid, line 28: IIUC, I would write this as `let candidates = orderedCandidates.filter(pol => pol.points > 0);`.

layouts/partials/blocks/candidate-results.html
```diff
@@ -0,0 +1,80 @@
...
+        const result = {};
+        for (const [key, value] of urlParams) {
+          result[key] = value;
+        }
+        return result;
+      }
+
+      let candidateScores = paramsToObject()
+
+      let orderedCandidates = candidateData.map(obj => ({
```
earthboundkid, line 20: This isn't sorted yet, so I would probably call it something like `candidateObjs` and maybe rename `candidates` to `sortedCandidates` or `rankedCandidates`.

content/ag-quiz.md
```diff
@@ -0,0 +1,33 @@
...
+
+The attorney general represents the interests of the state and the public, and defends Pennsylvania’s laws in court. The job has also served as a stepping stone for running for higher office in the commonwealth, including the governorship.
+
+This year, [five Democrats and two Republicans are running for the job](https://www.spotlightpa.org/news/2024/03/pennsylvania-election-2024-attorney-general-primary-candidates/). In many ways, the respective parties’ candidates are similar.
+
+All five Democratic candidates have vowed to defend abortion access and the state’s election results while promising to take a holistic approach to addressing violent crime. The two Republican candidates — both veterans who have been prosecutors — are running on tough-on-crime platforms and say the state constitution doesn’t guarantee a right to abortion.
+
+<div class="mt-9 md:-mx-20 shadow-md rounded-lg" data-tf-live="01HS8X1DTFCF3Q9KH0W69KG19W"></div><script src="//embed.typeform.com/next/embed.js?typeform-welcome=0"></script>
```
earthboundkid, line 33: I had to scroll to answer some of the questions and that confused the thingie and made it think I was skipping questions. Can we make the window taller?

layouts/shortcodes/answer-key/keirbradfordgrey.html
```diff
@@ -0,0 +1,40 @@
+{{ $answer := .Get "answer" }}
```
earthboundkid, line 1: Are this and layouts/shortcodes/answer-key/eugenedepasquale.html still supposed to be here or are they leftover from a draft version of the code?

layouts/shortcodes/answer-key/q.html
```diff
@@ -0,0 +1,32 @@
...
+        id: {{ $id }},
+        get expanded() {
+            return this.active === this.id
+        },
+        set expanded(value) {
+            this.active = value ? this.id : null
+        },
+    }"
+  role="region"
+>
```
earthboundkid, line 16: This will soon be doable in native HTML with no JS using `<details name="groupname">`, but it's not in Firefox yet, alas. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#name

layouts/shortcodes/answer-key/keirbradfordgrey.html
jeffrummel: These were leftovers and have been removed: https://github.com/spotlightpa/poor-richard/pull/604/commits/6f09b0d01086c57d6e976af86726b021cb75eb8e

layouts/_default/quiz-results.html
```diff
@@ -0,0 +1,127 @@
...
+          )
+        }}
+      </div>
+    {{ end }}
+    {{ with .Page.RenderString (.Param "answer-key") }}
+      <div
+        x-data="{ active: 1 }"
+        class="mx-auto max-w-2xl space-y-4 px-5 py-12 lg:px-0"
+      >
+        <h2 class="leading-tight; mb-4 text-3xl font-bold">Answer Key</h2>
```
earthboundkid, line 79: Scoot this a little closer to the "Meet the Candidates" button. I almost didn't see the key at first because it's further down the page and I expected there to just be story promos down there.

earthboundkid: Another solution is maybe to make "Answer Key" closer and colorful or decorated so it grabs the eye and you notice the key.

PR #617, opened 2024-04-01
Fix for pinned pages on series

earthboundkid:
Reverts spotlightpa/poor-richard#615

#615 removed pinned pages from series too, and the river isn't adjusted, so the river ended up not having the missing pinned articles. Let's revert and try it again.


earthboundkid:
The logic in the old page was wrong, so I ended up doing a new commit instead.

earthboundkid:
I added a separate topic.html as just a clone of list.html with the featured pages taken out. Longer term, we should probably deduplicate the HTML somewhat.

jeffrummel (approved):

PR #628, opened 2024-04-10
Jr/adds new partners

jeffrummel:


layouts/partials/blocks/partners-page.html
```diff
@@ -110,6 +110,7 @@
                 {{ range $partnerCounties }}
                   <option value="{{ . }}">{{ . }} County</option>
                 {{ end }}
+                <option value="Regional/At-large">Regional/At-large</option>
```
earthboundkid, line 113: The big list of counties has "Regional/At-Large" in the middle. Maybe change the loop above to

```html
                {{ range $partnerCounties }}
                   {{ if ne . "Regional/At-larg" }}
                     <option value="{{ . }}">{{ . }} County</option>
                   {{ end }}
                 {{ end }}
```

earthboundkid:
I noticed there's no dot in Susquehanna but it does have WSKG.

jeffrummel:
> I noticed there's no dot in Susquehanna but it does have WSKG.

WSKG is in New York state and covers the Susquehanna county region but no specific municipalities.

earthboundkid (approved):

PR #631, opened 2024-04-15
ballyhoo shortcode

jeffrummel:


layouts/shortcodes/featured/ballyhoo.html
```diff
@@ -0,0 +1,9 @@
+{{ partial "tw/ballyhoo.html" (dict
```
earthboundkid, line 1: Should we have this get the parameters from the short code attributes instead of hard coding them?

jeffrummel: I was thinking this would keep it simple since this is a one-off, but now I realize we should go ahead and make it complex in case we want to use the same shortcode to promote other stories/tools in the future -- https://github.com/spotlightpa/poor-richard/pull/631#discussion_r1565837714

earthboundkid (approved):

PR #633, opened 2024-04-16
swaps header on chatbot

jeffrummel:


layouts/_default/page-chatbot.html
```diff
@@ -0,0 +1,119 @@
...
+        class="flex-auto font-sans text-2xl font-extrabold capitalize leading-tight tracking-wide md:text-3xl"
+      >
+        {{- .Title -}}
+
+      </h2>
+      <div class="-order-1 w-full sm:order-1 sm:w-auto">
+        {{ partial "tw/breadcrumbs.html" . }}
+      </div>
+    </header>
+    {{ if or (.Content) (.Page.Params.subhed) }}
```
earthboundkid, line 30: This part is the same as in the regular page-landing. Maybe we should pull it out into its own partial.

jeffrummel: This pr is a hotfix. In the future I'd like to pull the header out into a shortcode similar to the [featured takeover ](https://www.spotlightpa.org/elections-2024/)

earthboundkid (approved):

PR #634, opened 2024-04-16
Custom Chart shortcode

jeffrummel:


layouts/shortcodes/custom/chart-item.html
```diff
@@ -0,0 +1,47 @@
+{{ $percent := .Get "percent" | default 0 }}
+{{ $inner := .Get "body" }}
+{{ $number := .Get "number" }}
+
+<style>.w-{{ $percent }} {width: {{ $percent }}%;}.max-w-{{ $percent }} {max-width: {{ $percent }}%;}</style>
```
earthboundkid, line 5: Does the autoformatter shrink it down like that? Can we expand it?

earthboundkid: Alternatively, you could have the style set CSS variables and then use them as classes.

This worked:

```html
<style>
  div {
    --my-var: 10%;
  }
</style>

<div class="w-[--my-var] bg-slate-500">
  hello
</div>
```

https://play.tailwindcss.com/ugndmjrMWZ

earthboundkid: Not tested, but this looks right:

```html
<figure
   x-data="{ onScreen: false }"
   x-intersect.half="onScreen = true"
   style="--item-pct: {{ $percent}}%;"
   class="mx-auto flex max-w-5xl flex-col justify-center py-6"
 >
  <div class="relative my-3 h-9 overflow-hidden rounded bg-yellow-7 md:h-12">
     <span
       x-show="onScreen"
       class="max-w-[--item-pct] block h-9 md:h-12"
```

earthboundkid: Oh yeah, besides aesthetics, the `<style>` tag solution won't work as is because if you have more than one chart-item on the page, later ones will override the style of earlier ones and all the items will come out the same. I think the variable solution will work though.

jeffrummel: I can't get either solution to work. If we apply a style tag with the var `style="max-width: --item-pct;"` it sets the max-width correctly, but `x-transition:enter-end="w-[--item-pct]"` overrides that because alpine and tailwind don't translate --item-pct into the percent.

earthboundkid: I think it's just not realizing it needs to keep the `w-[--item-pct]` class and stripping it from the CSS. Try adding `<div hidden class="w-[--item-pct]"></div>` at the bottom.

earthboundkid (approved):

PR #639, opened 2024-04-23
Add picture-rotator.html

earthboundkid:
In the course of this I realized that the header ad and topper ad are the same now, so we should just eliminate one.


jeffrummel (approved):

earthboundkid:
This improves first load time a bit when we're showing a top banner and prevents unnecessary loading of images for other screen sizes. Also it's 50 lines net negative! 🎉

PR #641, opened 2024-04-24
removes custom hed for chatbot

jeffrummel:


earthboundkid (approved):
Do you know when they want to take the password off?

jeffrummel:
> Do you know when they want to take the password off?

We're leaving the pw on for a while. At least until we can get the embed design updated.

PR #661, opened 2024-05-17
Article Redesign

jeffrummel:


earthboundkid:
image

I think that needs to be a little further from the nav bar. Also, do we still want to use the bird icon now that it's X?

PR #715, opened 2024-08-07
social-buttons: Fix broken X.com redirect

earthboundkid:


earthboundkid:
Looks like X fixed it.

PR #803, opened 2024-11-27
Add IDs to homepage, article page ads

earthboundkid:
Here are a couple more ad IDs that Christina wants for visibility tracking. Follow up to #799


layouts/partials/tw/promo-double.html
```diff
@@ -12,7 +12,11 @@
 }}
 
 
-<aside class="{{ $wrapperClass }}" data-ga-category="promo-image">
+<aside
+  class="{{ $wrapperClass }}"
+  data-ga-category="promo-image"
+  {{ with .id }}id="{{ . }}"{{ end }}
+>
```
earthboundkid, line 19: Not using this yet, but it seems like we should have it in case we do.

earthboundkid:
@alexandramharris In terms of not stepping on toes with PR merging, the rule is the person who created the PR is in charge of actually pushing the update and merge button. 

From [the blog post](https://blog.carlana.net/post/2018/git-gud/) about Git rules:

> Once the code has been positively reviewed and automated tests have passed, the author of the code should be the one to perform the actual merge operation. The reason for this is that inevitably, when someone else merges the code, there will be one more small thing that will jump out to the author just as the merge is about to happen. Having the author merge solves this problem and provides a simple convention as to whose job it is to do the merging.

Thanks!

PR #806, opened 2024-12-02
newsletter/hidden: Rotate shibboleth

earthboundkid:


earthboundkid:
We're having a bot surge, so I'm shooting from the hip here.

PR #813, opened 2025-01-10
tw/nav.html: Fix donate pop up on modal-excluded pages

earthboundkid:
If a page is 'modal-exclude', it won't have the Fundraise Up code, so the donate popup won't load.

### DIFF
--- a/layouts/partials/tw/nav.html
+++ b/layouts/partials/tw/nav.html
@@ -183,9 +183,11 @@
         data-fru-campaign="navbar"
         data-ga-category="donate"
         class="group flex items-center justify-around overflow-hidden rounded-md bg-gradient-to-r from-red-7 to-red p-2 text-white hover:from-red-5 hover:to-red-9 sm:justify-center md:-my-1 md:p-2 lg:px-4"
-        x-data
         href="/donate/"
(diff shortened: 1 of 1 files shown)


layouts/partials/tw/nav.html
```diff
@@ -183,9 +183,11 @@ <h3 class="font-bold uppercase tracking-wide hover:underline">
         data-fru-campaign="navbar"
         data-ga-category="donate"
         class="group flex items-center justify-around overflow-hidden rounded-md bg-gradient-to-r from-red-7 to-red p-2 text-white hover:from-red-5 hover:to-red-9 sm:justify-center md:-my-1 md:p-2 lg:px-4"
-        x-data
         href="/donate/"
-        :href="'#form=donate-onetime?utm_source=www.spotlightpa.org&utm_medium=home&utm_campaign=navbar'"
+        {{ if not (page.Param "modal-exclude") }}
+          x-data
+          :href="'#form=donate-onetime?utm_source=www.spotlightpa.org&utm_medium=home&utm_campaign=navbar'"
```
earthboundkid, line 189: `x-data` and `:href` are https://alpinejs.dev/ code. If a user has JavaScript, it will change the href to the string, but if they don't have JS, the default href will still be there.

alexandramharris (approved):

PR #824, opened 2025-01-29
Ah/awards

alexandramharris:
Adding the impact page to the site. Consists of an intro section, story carousel, explanation for tracking impact, impact by the numbers and list of awards by year.


content/about/_index.md
```diff
@@ -83,6 +83,9 @@ actionText = "Our Team"
 url = "/about/board/"
 actionText = "Board of Directors"
 
+[[blocks.cta]]
+url = "/about/impact/"
```
earthboundkid, line 87: The URL is /impact/, but maybe it should be /about/impact/. Either way, probably it makes more sense to have the actual file be content/about/impact.md just so it's with the other marketing pages.

layouts/partials/blocks/carousel.html
```diff
@@ -0,0 +1,522 @@
+<div id="carousel-container" class="carousel-container pt-16">
```
earthboundkid, line 1: When I opened the page, it took a second for the image to load, so it looked wrong against a white background. Maybe give put a light gray background under the carousel for before it loads?

data/carousel.toml
```diff
@@ -0,0 +1,72 @@
+[[items]]
+title = "Spotlight PA Featured Impact"
+tagline = "Housing"
+text = "A Spotlight PA investigation into Pennsylvania’s pandemic mortgage relief program found that long delays and poor communication were leaving thousands in limbo, with their homes and utilities on the line. The story prompted the state to fire the third-party contractor that was running the program."
+short = "The state fired a third-party contractor after Spotlight PA revealed serious issues with Pennsylvania’s pandemic mortgage relief program."
+link = "https://www.spotlightpa.org/series/the-waiting-game/"
+linkText = "Read the series"
+image = "external/s36w4vxgrg2nnpxjrf10fce26m.jpeg"
+width = "2400"
```
earthboundkid, line 9: I think it probably makes more sense to leave the sizes out of the data and just have all of the carousel images resized to something reasonable like 1200x800.

data/carousel.toml
```diff
@@ -0,0 +1,72 @@
+[[items]]
+title = "Spotlight PA Featured Impact"
+tagline = "Housing"
+text = "A Spotlight PA investigation into Pennsylvania’s pandemic mortgage relief program found that long delays and poor communication were leaving thousands in limbo, with their homes and utilities on the line. The story prompted the state to fire the third-party contractor that was running the program."
+short = "The state fired a third-party contractor after Spotlight PA revealed serious issues with Pennsylvania’s pandemic mortgage relief program."
+link = "https://www.spotlightpa.org/series/the-waiting-game/"
```
earthboundkid, line 6: Make links relative so local dev/a deploy preview doesn't accidentally send you to the real site and then you wonder why it isn't refreshing.

layouts/_default/impact.html
```diff
@@ -0,0 +1,83 @@
+{{ define "main" }}
```
earthboundkid, line 1: This is mostly the same as the list layout, right? I'd be interested in seeing if we could switch this to just using blocks and deleting impact.html as a follow up PR later. It doesn't need to be done now though.

layouts/_default/impact.html
```diff
@@ -0,0 +1,83 @@
...
+    <div
+      data-ga-category="content"
+      class="article-content mx-auto w-full px-4 py-6 pb-10 md:w-7/12 md:px-0"
+    >
+      {{ if .Params.intro }}
+        {{ .Params.intro | markdownify }}
+      {{ end }}
+    </div>
+
+    <div id="award-quote-block" class="mx-auto md:w-7/12 md:px-0">
```
earthboundkid, line 21: Avoid using IDs because it means we can't ever repeat something on the page later. A lot of things you think you only need one of end up needing to be repeated. For things that do require IDs, like label-input pairs, we have a template helper:

```
{{ $id := partial "helper/new-id" . }}
<label for="{{ $id }}>
  Something
</label>

<input id="{{ $id }}" />
```

It just automatically increments label-uid-1, label-uid-2, etc. each time you use it on a page, so each call to the partial is unique.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,269 @@
...
+      </svg>
+      <h2 class="text-4xl font-bold">Awards</h2>
+    </div>
+
+    <div class="mx-auto max-w-4xl px-4">
+      <div style="border-top: 2.5px solid black; width: 100%;"></div>
+
+      <div class="flex w-full">
+        <div
+          id="years-padding"
```
earthboundkid, line 15: Lets do this styling as Tailwind classes instead.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,269 @@
...
+
+            {{ range $index := seq 0 (sub $count 1) }}
+              <li>{{ index $sortedYears (sub (sub $count 1) $index) }}</li>
+            {{ end }}
+          </ul>
+        </div>
+
+        {{ $awards := site.Data.awards }}
+        <script>
+          document.addEventListener("DOMContentLoaded", () => {
```
earthboundkid, line 35: Doesn't need to be rewritten now, but there's probably a simpler way to write this with Alpine.js. Probably just put an `x-data="{ activeYear: '2024' }"` on the container and then `x-show="$el.dataset.year === activeYear" data-year="{{ .year }}"` on the main panel and `@click="activeEl = $el.dataset.year" data-year="{{ .year }}"` on the sidebar links.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,269 @@
...
+              {{ range where (where . "national_or_state" "National") "winner_or_finalist_honorable_mention" "Winner" }}
+                <div class="award-entity" data-year="{{ .year }}">
+                  {{ .entity }}:
+                  <span class="award-category">{{ .award_category }}</span>
+                  <p class="award-story">
+                    {{ if .link }}
+                      <a
+                        href="{{ .link }}"
+                        class="award-story-link"
+                        target="_blank"
```
earthboundkid, line 117: We actually have some magic JS on the site that makes all off site links open in a new window, so we probably don't need target="_blank" here.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,269 @@
...
+            <p class="award-national-state national-awards-header">
+              National Awards
+            </p>
+            <div class="national-awards">
+              {{ range where (where . "national_or_state" "National") "winner_or_finalist_honorable_mention" "Winner" }}
+                <div class="award-entity" data-year="{{ .year }}">
+                  {{ .entity }}:
+                  <span class="award-category">{{ .award_category }}</span>
+                  <p class="award-story">
+                    {{ if .link }}
```
earthboundkid, line 113: We have a helper partial because this comes up a lot: 

```
{{ partial "tag/maybe-link.html" (dict
      "link" .link
      "text" .story
      "class" "award-story-link"
  )
}}
```

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,269 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+{{ end }}
+
+
+<style>
```
earthboundkid, line 213: I'd like to get these turned into Tailwind. Instead of doing the first-child/last-child margin 0 and having to use 7.5px of margin, it would probably be easier to have `flex flex-col gap-[15px]`.

layouts/partials/blocks/carousel.html
```diff
@@ -0,0 +1,522 @@
...
+      if (scrollLeft + clientWidth < scrollWidth) {
+        rightGradient.style.display = "block";
+      } else {
+        rightGradient.style.display = "none";
+      }
+    }
+
+    bottomBarContainer.addEventListener("scroll", updateGradients);
+
+    window.addEventListener("resize", updateGradients);
```
earthboundkid, line 208: These should probably be debounced so we don't hurt resize/scroll performance.

layouts/partials/blocks/carousel.html
```diff
@@ -0,0 +1,522 @@
...
+
+    updateCarousel();
+    updateGradients();
+  });
+</script>
+
+<style>
```
earthboundkid, line 215: This seems like it's probably sufficiently complex that it's not worth translating into Tailwind, but we can move it into assets/css/carousel.css and link it from there just to simplify the page a little.

```
{{ $css := resources.Get "css/carousel.css" | minify | fingerprint }}
<link rel="stylesheet" href="{{ $css.RelPermalink }}" />
```

earthboundkid (commented):
Code looks pretty good. I'll take a closer look at the design tomorrow.

layouts/partials/blocks/carousel.html
alexandramharris: A side note, how would we feel about changing the mobile carousel to cycle through the stories by scrolling instead of clicking the buttons? It feels like that might be a better user experience — maybe in a future update?

layouts/partials/blocks/awards.html
earthboundkid: Oh, this should replace the `{{ if .link }}` and `{{ else }}` code. It's just a shortcut for the "if .link use an `a`, else just dump the text" pattern.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,222 @@
...
+            </p>
+            <div class="national-finalists">
+              {{ range where (where . "national_or_state" "National") "winner_or_finalist_honorable_mention" "Finalist/Honorable Mention" }}
+                <div class="award-entity" data-year="{{ .year }}">
+                  {{ .entity }}:
+                  <span class="award-category font-bold"
+                    >{{ .award_category }}</span
+                  >
+                  <p class="award-story mb-5">
+                    {{ if .link }}
```
earthboundkid, line 147: Same here.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,222 @@
...
+              class="award-national-state national-finalists-header pb-[10px] text-[22px] font-bold"
+            >
+              National Finalists
+            </p>
+            <div class="national-finalists">
+              {{ range where (where . "national_or_state" "National") "winner_or_finalist_honorable_mention" "Finalist/Honorable Mention" }}
+                <div class="award-entity" data-year="{{ .year }}">
+                  {{ .entity }}:
+                  <span class="award-category font-bold"
+                    >{{ .award_category }}</span
```
earthboundkid, line 144: I find this wrapping ugly, so usually I do

```
<span class="award-category font-bold">
    {{- .award_category -}}
</span>
```

The `{{-` and `-}}` trim the whitespace between the tag and template code. The usually only matters for links where if you don't do it, the underline is too long because it's underlining a space.

layouts/partials/blocks/awards.html
```diff
@@ -0,0 +1,222 @@
...
+            </p>
+            <div class="state-finalists">
+              {{ range where (where . "national_or_state" "State") "winner_or_finalist_honorable_mention" "Finalist/Honorable Mention" }}
+                <div class="award-entity" data-year="{{ .year }}">
+                  {{ .entity }}:
+                  <span class="award-category font-bold"
+                    >{{ .award_category }}</span
+                  >
+                  <p class="award-story mb-5">
+                    {{ if .link }}
```
earthboundkid, line 203: Here too.

layouts/partials/blocks/carousel.html
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      if (scrollLeft + clientWidth < scrollWidth) {
+        rightGradient.style.display = "block";
+      } else {
+        rightGradient.style.display = "none";
+      }
+    }
+
+    const debouncedUpdateGradients = debounce(updateGradients, 20);
```
earthboundkid, line 221: I like to put in a comment so you know what the units for a timeout are. `const debouncedUpdateGradients = debounce(updateGradients, 20 /* ms */);` The other way to do it is to call the function `debounceMS`.

layouts/partials/blocks/by-the-numbers.html
```diff
@@ -0,0 +1,494 @@
...
+          </div>
+        </div>
+        <div class="impact-row symbol-number">1,400+</div>
+        <div class="impact-row impact-text">Right to Know Requests</div>
+      </div>
+    </div>
+  </div>
+</div>
+
+<style>
```
earthboundkid, line 72: I do want to take the time to convert this to Tailwind, just because we're using Tailwind everywhere else.

I found a site https://transform.tools/css-to-tailwind that will give us a first pass at the conversion. For each rule, you'd just find the Tailwind equivalent and then substitute that for the class name. So, `id="by-the-numbers-wrapper` would be `class="font-serif"` `id="numbers-bar"` would be `class="bg-yellow h-[110px] flex w-full items-center` etc. The media query ones will be a little harder because Tailwind classes start as mobile and change as you get larger instead of starting as desktop and changing as you get smaller, but it should be doable.

```css

**Review by earthboundkid — COMMENTED** [review]
Here are some more changes.

**Comment by earthboundkid** [comment]
> Here are some more changes.

Oh, my bad, I thought I was doing a joint review comment bomb, but it was sending each comment out individually.

File: layouts/partials/blocks/carousel.html (hunk 12 above)
Comment (earthboundkid) [reply layouts/partials/blocks/carousel.html]: I would expect an auto-rotate until the first interaction, maybe. Then it becomes manually controlled.

File: layouts/partials/blocks/carousel.html (hunk 17 above)
Comment (alexandramharris) [reply layouts/partials/blocks/carousel.html]: Maybe an idea for a future PR: Thoughts on changing the carousel bottom grey bar from a gradient system to something that uses a scroll progress bar? I find the gradient and "active" state for which story is currently clicked (where it goes semi-transparent and grey) somewhat difficult to understand on certain screen sizes. A progress bar might be clearer and a better UX.

File: layouts/partials/blocks/awards.html [hunk 19]
```diff
@@ -0,0 +1,211 @@
...
+              {{ $years = $years | append .year }}
+            {{ end }}
+            {{ $sortedYears := uniq $years | sort }}
+            {{ $count := len $sortedYears }}
+
+            {{ range $index := seq 0 (sub $count 1) }}
+              <li class="year-link cursor-pointer hover:text-black">
+                {{ index $sortedYears (sub (sub $count 1) $index) }}
+              </li>
+            {{ end }}
```
earthboundkid, line 35: IIUC, I think you can do this instead and it's more clear:

```html
{{ $years := slice }}
{{ range . }}
  {{ $years = $years | append .year }}
{{ end }}
{{ $yearsDesc := sort (uniq $years) "value" "desc" }}
{{ range $year := $yearsDesc }}
  <li>
    <button class="year-link hover:text-black" type="button">
      {{ $year }}
    </button>
  </li>
{{ end }}
```

Using a button instead of a plain LI gets you a free cursor hand and also helps screen readers know that these are clickable.

earthboundkid (commented):
Other than the li button thing, I think we can launch this now and go back and make other tweaks when we get around to it. Congrats, looks great. 🎉

earthboundkid (approved):

PR #833, opened 2025-02-11
Ah/award links

alexandramharris:


layouts/partials/blocks/awards.html
```diff
@@ -121,10 +121,15 @@ <h2 class="text-4xl font-bold lg:text-5xl">Awards</h2>
                     {{- .award_category -}}
                   </span>
                   <p class="award-story mb-5">
+                    {{ $class := "" }}
+                    {{ if and (isset . "link") (ne .link "") }}
+                      {{ $class = "award-story-link text-[#22416e] underline transition-colors duration-500 ease-linear hover:text-[#ffcb03]" }}
+                    {{ end }}
+
                     {{ partial "tag/maybe-link.html" (dict
+                      "class" $class
```
earthboundkid, line 130: You can drop the `if` and just use `"linkClass" "award-story-link text-[#22416e] underline transition-colors duration-500 ease-linear hover:text-[#ffcb03]"`. The `linkClass` attribute isn't in your branch, but it was added to `maybe-link.html` on master a few weeks back.

earthboundkid: https://github.com/spotlightpa/poor-richard/blob/master/layouts/partials/tag/maybe-link.html Looks like the `linkClass` prop was added as part of doing the Leaders in Action page three weeks ago.

earthboundkid (approved):

PR #847, opened 2025-02-26
Remove duplicate award

alexandramharris:
Waiting to confirm which description we want before merging

### DIFF
--- a/data/awards.json
+++ b/data/awards.json
@@ -287,19 +287,6 @@
   },
   {
     "year": "2024",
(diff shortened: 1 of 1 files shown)


alexandramharris:
Confirmed we want the description:

> For "Richest Little City," an investigation into how the manager of a small Pennsylvania city landed at the center of one of the most sweeping political corruption cases in the state in recent years.

earthboundkid (approved):

PR #855, opened 2025-03-05
Hide sticky mobile contribute popup

alexandramharris:


layouts/partials/tw/frontpage.html
```diff
@@ -119,7 +119,7 @@ <h2 class="text-2xl font-extrabold">
     <div
       data-ga-category="midpage"
       class="bottom-0 row-start-4 md:relative md:col-span-9 md:mx-0 md:overflow-hidden md:rounded xl:col-span-10"
-      :class="isSticky ? 'sticky -mx-4 bottom-0 md:mx-0' : 'relative mx-0 rounded overflow-hidden'"
+      :class="isSticky ? 'hidden sticky -mx-4 bottom-0 md:mx-0' : 'relative mx-0 rounded overflow-hidden'"
```
earthboundkid, line 122: We can just delete all this and change it to include newsletter/papost-hidden.html instead.

earthboundkid:
image

The HP newsletter box is still sticky on mobile.

image

The X to close the sticky donate box is still there, but the actual donate box is hidden.

I'd prefer to just make them both non-sticky and lose the X, since they won't need to be closed once they're not sticky.

earthboundkid (approved):

PR #885, opened 2025-04-24
Move Berks County pages into their own subsection

earthboundkid:


layouts/berks/single.html
```diff
@@ -0,0 +1,9 @@
+{{ define "body-class" }}font-serif{{ end }}
+
+{{ define "body-attributes" }}
+  data-fru-campaign="state-college-navbar" data-fru-theme="sc-onetime"
```
earthboundkid, line 4: FIX THIS

alexandramharris (approved):

PR #896, opened 2025-05-05
Article Redesign

alexandramharris:


content/news/2024-05-14-SPLGAZAORDER.md
```diff
@@ -27,7 +27,7 @@ The vaguely worded update, which went into effect last week without the public f
 
 It has also heightened fears among pro-Palestinian and Muslim American groups that employees who express opinions that differ from the Democratic governor’s stance on the war could be unfairly punished.
 
-{{<embed/newsletter>}}
+<script src="https://www.spotlightpa.org/embed.js" async></script><div class="spl-newsletter" data-spl-embed-version="1" data-spl-src="https://www.spotlightpa.org/embeds/newsletter/"></div>
```
earthboundkid, line 30: Can we just apply `spl-newsletter` to a container in embed/newsletter? I don't want to stop using shortcodes for newsletters. It makes it possible to swap them out for links in the app, which is an important goal.

layouts/partials/tw/related-stories.html
earthboundkid, line 1: This is blank.

alexandramharris: I think [Jeff's was]( https://github.com/spotlightpa/poor-richard/blob/11e3b4cfac504ad0f25c405991bf8bd0d28338b1/layouts/partials/tw/related-stories.html) as well?

Do we know if this was ever coded or designed? If we have a design, I could code that, or chat about it and start from scratch if we want.

layouts/embed-page/single.html
```diff
@@ -19,7 +19,7 @@
 {{ end }}
 
 {{ define "main" }}
-  <div class="bg-white p-px" data-ga-category="embed" data-fru-campaign="embed">
+  <div class="bg-g-2 p-px" data-ga-category="embed" data-fru-campaign="embed">
```
earthboundkid, line 22: Revert the embed changes on this PR and just do them in the follow up PR.

assets/css/tw.css
```diff
@@ -253,6 +253,13 @@
   @apply inline-block rounded border-2 border-blue bg-blue px-6 py-2 text-center font-sans font-semibold uppercase tracking-widest text-white no-underline outline-none ring-blue hover:bg-blue-7 focus:ring-2 focus:ring-blue active:ring-blue;
 }
 
+.spl-newsletter {
+  @apply my-2 sm:float-right sm:ml-3 sm:w-1/2 md:-mr-12 md:max-w-[439px] lg:-mr-0 lg:w-full lg:max-w-96 xl:-mr-20 xl:max-w-[475px];
+}
+.spl-donate {
+  @apply my-2 sm:float-left sm:mr-3 sm:w-1/2 md:max-w-[439px] lg:w-full lg:max-w-96;
+}
```
earthboundkid, line 261: These two changes are really part of the embed redesign, so lets drop them from here and just make this a clean change to just the article page itself.

alexandramharris: Agreed, I've moved some of them over already but I can make sure to revert them here as well.

layouts/partials/tw/main.html
```diff
@@ -30,225 +16,242 @@
...
+    {{ end }}
 
-      <div class="{{ $mainWellClass }}">
-        <div x-data x-intersect.once="$dispatch('open-sticky')"></div>
-        {{ partial "tw/content-well.html" . }}
+  </header>
+  <article
+    class="mx-auto flex max-w-screen-md flex-wrap md:justify-center md:gap-x-12 lg:col-span-8 lg:justify-start lg:gap-x-8"
+  >
+    <div class="w-full print:!max-w-md">
```
earthboundkid, line 95: I don't think we need `print:!max-w-md` anymore.

Let's add a todo item to check the print styles in a follow up PR.

layouts/partials/tw/main.html
```diff
@@ -1,20 +1,6 @@
 {{ $params := partialCached "helper/page-params" .Page .RelPermalink }}
 {{ $hasImage := .Param "image-size" | ne "hidden" }}
 {{ $isWide := .Param "image-size" | eq "wide" }}
-{{ $isOneCol := (not $hasImage | or $isWide) }}
```
earthboundkid, line 4: The old page had this logic to figure out if the page should be one column. Hiding the main image still works, but the image-size = wide doesn't work anymore. I think we can delay doing this, but I do want it to be a todo item to bring it back because some articles really do need to drop the right rail.

earthboundkid (approved):
I think we can merge this, but I want us to do follow ups for:

- The embeds (mostly ready)
- Print styles (shouldn't need much change)
- Wide layouts

PR #901, opened 2025-05-13
Ah/embed redesign

alexandramharris:


layouts/partials/tw/content-well.html
```diff
@@ -1,9 +1,12 @@
...
     <div class="mt-6" data-ga-category="content" id="content">
-      {{ partial "tag/content-well-start" . }}
-      {{ . }}
-      {{ partial "tag/content-well-end" . }}
+      <div
+        class="article-content w-full leading-relaxed"
+        data-sf-campaign="701Dn000000YgozIAC"
+      >
+        {{ . }}
+      </div>
```
earthboundkid, line 9: This probably should not change. The content-well-start tag has `article-content w-full overflow-hidden leading-relaxed`. Not sure if we want overflow-hidden or not, but probably. The data-sf-campaign is just outdated. When we switched to Fundraise Up from Salesforce, we changed all these to be data-fru so that definitely shouldn't change.

earthboundkid:
The Github diff is showing a bunch of changes that are really from the article page change. Can you merge master so it only shows these changes?

layouts/embed-page/single.html
```diff
@@ -19,7 +19,7 @@
 {{ end }}
 
 {{ define "main" }}
-  <div class="bg-white p-px" data-ga-category="embed" data-fru-campaign="embed">
+  <div class="bg-g-2 p-px" data-ga-category="embed" data-fru-campaign="embed">
```
earthboundkid, line 22: This should be white still, because it needs to blend in on other people's article wells. The inside of the embed can be gray 2 though.

earthboundkid:
Can you add a page to https://www.spotlightpa.org/design/ that just demonstrates all the shortcodes? It would be useful for previewing this, and probably we'll want it when whenever we revise them.

layouts/partials/tw/content-well.html
alexandramharris: When we keep `overflow-hidden`, we end up cutting off some of the floating embeds. I currently removed it, but if we need it for other reasons, it might just mean refactoring the code further to not rely on removing it. I'm not noticing any negative impacts of removing it, but if there's any special pages we need to consider, we can.

earthboundkid:
Can we tweak these margins and play with the line height a little? Put the horizontal line closer to the hed, then the dek further from the line, tighten the line height a little. The red button should be further away from the text, but sign up should be closer to the email address. Maybe make the email+button one line on desktop and two on mobile.

earthboundkid (commented):
Very close. The actual embeds/shortcodes all look great!

PR #908, opened 2025-05-29
Shortcodes: Add clear-both to embeds

earthboundkid:
To help them clear any floating embeds


alexandramharris (approved):
Thank you 🙏 Just getting around to this, but I can delete my branch. This looks good.

PR #914, opened 2025-06-03
Remove HWC from menu

alexandramharris:


alexandramharris:
I think this just leaves embeds like this: https://deploy-preview-914--poor-richard-spotlightpa.netlify.app/news/2025/05/private-equity-crozer-hospitals-philadelphia-report-pennsylvania/

Is this something we can handle in BlueLena, or better to do a find and replace for "Get the latest on the state of caregiving in Pennsylvania with our free weekly newsletter, How We Care"?

earthboundkid (commented):
I don't want to delete the old files for newsletters per se because we don't need those to stop working and it would be better if you could still read them. Maybe we can just make the index page not render somehow and then have the plain newsletters/howwecare/ redirect?

data/redirects.toml
```diff
@@ -1,3 +1,15 @@
...
+from   = "/newsletters/howwecare"
+to     = "/newsletters"
+status = 301
+force  = true
+
+[[redirects]]
+from   = "/newsletters/howwecare/"
+to     = "/newsletters"
+status = 301
+force  = true
```
earthboundkid, line 11: I don’t think you need both. It should get the slash and no slash versions with one redirect.

earthboundkid (approved):

PR #921, opened 2025-06-17
Adding class for external developer to target and hide sidebars in app

alexandramharris:


earthboundkid:
In the app, they should be seeing the .json variations on the shortcodes instead, so this shouldn't be needed/show up. Not sure what's up. Maybe something is going wrong or they're pulling from the live URL and not the feed.

PR #937, opened 2025-07-21
Update list of partners

earthboundkid:


alexandramharris:
I think this is maybe a context issue:

```
"partials/blocks/partners-page.html" at <.County>: can't evaluate field County in type interface {}
```

PR #940, opened 2025-07-23
Ah/scrolly

alexandramharris:


earthboundkid (approved):
Impressive!

layouts/shortcodes/featured/scrolly.html
```diff
@@ -1,13 +1,53 @@
...
+     } else {
+       this.buttonPosition = 'fixed';
+     }
+   }
+ }"
+  x-init="
+   $nextTick(() => {
+     updateButtonPosition();
+     window.addEventListener('scroll', () => updateButtonPosition());
+     window.addEventListener('resize', () => updateButtonPosition());
```
earthboundkid, line 26: These need to be debounced. Otherwise you're calculating BoundingClientRect for every sub-pixel of scrolling and it slows the computer down too much.

earthboundkid: You could do it by using `@scroll.window.debounce="updateButtonPosition()"` instead, I think.

earthboundkid (approved):

PR #942, opened 2025-07-24
Add new option to audio player

alexandramharris:


layouts/shortcodes/audio.html
```diff
@@ -1,17 +1,39 @@
 {{ $label := .Get "label" | default `Click here to listen to an audio version of this story.` }}
 {{ $src := .Get "src" }}
+{{ $open := .Get "open" | default "false" }}
+{{ $isOpen := eq (lower $open) "true" }}
 
+{{ if not $isOpen }}
```
earthboundkid, line 6: Reverse the branches and do `{{ if $isOpen }}` here.

layouts/shortcodes/audio.html
```diff
@@ -1,17 +1,39 @@
...
+    </audio>
+  </figure>
+{{ else }}
 
-<figure class="clear-both" x-data="{ isOpen: false }">
-  <figcaption>
-    <button
-      class="cursor-pointer italic text-navy underline"
-      @click.once="isOpen = true; $refs.player.play()"
+  <figure class="clear-both w-full" x-data="{ isOpen: true }">
```
earthboundkid, line 22: For the isOpen case, we can just drop `x-data="{ isOpen: true }`, `x-show="isOpen"`, and `x-ref="player"` since we're just always showing it and using the native controls.

PR #948, opened 2025-07-31
Embeds: Add dropdown for newsletters

earthboundkid:


alexandramharris:
Looks good! Should we style the dropdown more like a dropdown than an input text field? Maybe just adding a down arrow or something simple.

PR #949, opened 2025-08-01
Ah/listening tour

alexandramharris:


earthboundkid:
image

Can you turn off the top ad? Or fix the margins on it? The page as a whole is kind of an ad anyway, so probably easiest to turn off.

earthboundkid (approved):

PR #951, opened 2025-08-01
Embeds: Fix missing papost default

earthboundkid:


earthboundkid:
I'm going to merge this before review just so no one else gets signed up wrong.

PR #995, opened 2025-10-16
Ah/inspections

alexandramharris:


layouts/_default/restaurant-inspections.html
```diff
@@ -0,0 +1,2196 @@
...
+            </div>
+          </div>
+
+          <button
+            @click="
+                    const cf = document.getElementById('cuisine-filter'); if (cf) cf.value = '';
+                    document.getElementById('city-filter').value = '';
+                    document.getElementById('date-sort').selectedIndex = 0;
+                    document.getElementById('risk-sort').selectedIndex = 0;
+                    document.getElementById('name-sort').selectedIndex = 0;
```
earthboundkid: Instead of using `getElementById`, you can use Alpine's data x-model to store input values in a JS field. https://alpinejs.dev/directives/model Also in Alpine you can look up elements with $ref instead of using an ID, which lets you have multiple of the same thing on a page (doesn't matter in this case but useful). https://alpinejs.dev/magics/refs In this case though, you'd probably want to just store the sortBy fields as x-models.

earthboundkid (commented):
Okay there's a lot here, more than I can review at once, but let's start by pulling all the JS into x-data functions, like how the ones in enhancements.js work. Then we can pull out the little utilities, like levenstein distance and htmlEscape, into the utilities folder and import them.

 Longer term but not right now, I would like to have the UI and data model pulled apart, so it's more maintainable. Right now, all the UI and logic are commingled in a way that makes it hard to tell what's going on.

So, instead of using `getElementById`, you would use Alpine's data x-model to store input values in a JS field. https://alpinejs.dev/directives/model Also in Alpine you can look up elements with $ref instead of using an ID, which lets you have multiple of the same thing on a page (doesn't matter in this case but useful). https://alpinejs.dev/magics/refs In this case though, you'd probably want to just store the sortBy fields as x-models.

So instead of 

```
<input id="this-one">
thing = document.getElementById("this-one") * 2 + 1;
```

You do things like

```
<div x-data="{thisOne: ''}">
  <input x-model="thisOne">

thing = this.thisOne * 2 + 1;
```

That's a big change though, so it can wait. I do think it will be worth it though if we want this project to be around a long time.

earthboundkid (approved):
🎉 🎉 🎉 🎉

PR #1110, opened 2026-04-15
Hide for NA

alexandramharris:


earthboundkid (approved):
This works. I think longer term we should rework the header to be more flexible. It could have sub-shortcodes for the various titles and whatnot. We don't want to break all the old pages though so maybe it would need to be a new thing.

PR #1111, opened 2026-04-15
Add stacked header option, top or bottom image option

alexandramharris:
This adds the option for editors to use these in featured stories:
```
  stack-layout="true"
  image-order="bottom" or "top"
```
For header images that we don't want to cut off on desktop.


alexandramharris:
Also adding a new option to our HR

earthboundkid (commented):
For the `.Get` calls, I like them all at the top of the file, so you can see what they all are at a glance without having to read the whole thing. Also maybe it should be `layout="stack"` rather than `stack-layout=true`.

layouts/_shortcodes/featured/hr.html
```diff
@@ -1,7 +1,16 @@
 <!-- wrapper for content-area -->
 {{ $short := .Get "short" }}
+{{ $style := .Get "style" | default "yellow" }}
```
earthboundkid: I would expect `style` to set the `style=` attribute on the container. Maybe this should be called `layout` or `kind` instead?

earthboundkid: We haven't done many short HRs, so we could just find them all and replace them with `kind=short` and this can be `kind=dot`.

layouts/_shortcodes/featured/header.html
```diff
@@ -50,10 +51,16 @@
 
 
 <header
-  class="flex flex-col md:min-h-screen md:flex-row"
+  class="{{ if ne $layout "stack" }}
```
earthboundkid: Having quotes inside the class confuses the syntax highlighter. I would either use backticks like 

    class="{{ if ne $layout `stack` }}

Or better just pull the logic out for this and the ones below and define the classes at the top of the file then do `class="{{ $headerClass }}"` `class="{{ $textOverlayClass }}"` etc.

earthboundkid (approved):

PR #1114, opened 2026-04-16
Add spacing

alexandramharris:
Editors requested spacing between these modules. Okay to add to all uses of this module, or do we need to do it conditionally?

image


earthboundkid (approved):
I think it's okay to just turn on for now. If we find some place that shouldn't have it, we can go back and tweak it again.

PR #1116, opened 2026-04-17
Move logo for images on top

alexandramharris:


earthboundkid (commented):
Ah, I see the issue here. The logo being on the featured template instead of in the featured/header shortcode creates various problems, but we don't want to go back and break all existing templates. I like the idea of how you're solving it, but let's use a different name for the parameter. It can be `hide-logo = true` on the page because it's not really about the "image" per se as much as it's about the logo. The alternative would be to reuse the `image-size` parameter, which normal pages have but featured pages don't use, but again, I don't think that's very descriptive because this is an issue with the logo placement not the image.

Another idea would be to have the featured template do 

```
  {{ if not (page.HasShortcode "featured/logo") }}
      <!-- existing code for the logo -->
```

And then make up a featured/logo shortcode that we use to put the logo somewhere else on the page. Maybe we can do that if/when we rework the header.

PR #1129, opened 2026-05-05
Name topics/series consistently

earthboundkid:
This branch screws up git on macOS because it renames files based on file case. The solution is to either checkout each commit in order to get past it or just delete your folder and clone a new one.


alexandramharris:
Looks good. Should this one have an image? https://deploy-preview-1129--poor-richard-spotlightpa.netlify.app/series/highway-stop-and-frisk/

alexandramharris (approved):

PR #1134, opened 2026-05-12
Ah/rest updates

alexandramharris:


netlify/functions/sync-ac.js
```diff
@@ -0,0 +1,143 @@
...
+      list: AC_LIST_ID,
+      contact: contactId,
+      status: 2,
+    },
+  });
+}
+
+export const handler = async () => {
+  try {
+    const subscribers = await getAllDynamoSubscribers();
```
earthboundkid: Won't this just get slower and slower we get more and more subscribers?

earthboundkid (approved):

PR #1150, opened 2026-06-01
Upgrades for June

earthboundkid:
Didn't do the Stripe upgrades because I don't know how to test it.


alexandramharris:
For Stripe, we'd need to loop Sean in for login access to check that it's receiving the subscriptions and also potentially test promo codes.

alexandramharris (approved):

PR #1154, opened 2026-06-05
scroll to top of results on mobile pagination

alexandramharris:


src/enhancements/inspections-data.js
```diff
@@ -488,11 +488,24 @@ export default function inspectionsData() {
...
+            scrollTarget.scrollIntoView({ behavior: "smooth", block: "start" });
+          }
+        } else {
+          const scrollTarget = document.querySelector(".inspections-mount");
+          if (scrollTarget) {
+            const top =
+              scrollTarget.getBoundingClientRect().top +
+              window.pageYOffset -
+              73;
+            window.scrollTo({ top, behavior: "smooth" });
```
earthboundkid: Is there a reason this one is using `window.scrollTo` instead of just making `.inspections-mount` the default target for `scrollTarget.scrollIntoView`  if `scrollTarget` is empty?

earthboundkid (approved):

PR #1155, opened 2026-06-05
defer test for Safari

alexandramharris:


earthboundkid (commented):
Is this a usecase for https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide ?

earthboundkid (approved):

PR #1167, opened 2026-06-30
Access Harrisburg: Link to Beehiv

earthboundkid:
We'll need a separate PR to go through and remove all the code that is obsolete now.


alexandramharris:
Do you think we should just link to it here since the monthly and annual buttons are kind of obsolete now?

Screenshot 2026-06-30 at 10 24 33 AM

earthboundkid:
The change to override that link is more complicated, so I'll leave it to you in a follow up PR.

PR #1173, opened 2026-07-06
add redirect

alexandramharris:


earthboundkid (approved):
Do we need to delete associated pages too?

alexandramharris:
> Do we need to delete associated pages too?

Like as part of this request or the larger refactor? 

Screenshot 2026-07-06 at 1 28 09 PM

earthboundkid:
Yeah as like a follow on PR to get rid of the obsolete stuff. This should be good on its own though.

PR #1180, opened 2026-07-15
Ah/colin requests

alexandramharris:


earthboundkid (approved):
I didn't know we had a magic eyebrow for eventDate. We should document that somehow.

PR #1198, opened 2026-08-17
hide support

alexandramharris:


earthboundkid (approved):
This isn't very elegant, but I don't think there's a better way to do it since you can't cascade from topic down to page. Maybe if there's demand for a second exclusion we can add a param onto the topic page and check page.topic.param.hide-whatever but this works for now.

PR #1203, opened 2026-08-27
Ah/individual games

alexandramharris:
Still need:

- Copy descriptions for each game page
- Check the Games email is set on backend
- Which games to feature on each page's "More games" section
- Do we want it to be `/word-flower-games/` or `/games/word-flower-games/`
- Inline newsletter success message copy (may want Christina's take on wording). The current thanks page doesn't really work for games (You’re about to get the best journalism) and also takes the user away from their game progress: https://www.spotlightpa.org/newsletters/thanks/
- Leaderboard is removed for now; API requires a higher-tiered plan than we have, testing out streaks available to us through our Pro plan

Everything else should be set, will check with Christina on these items upon return


layouts/_partials/newsletter/games-hidden.html
```diff
@@ -0,0 +1,2 @@
+{{ partial "newsletter/basic-hidden.html" . }}
+<input type="hidden" name="games" value="1" />
```
earthboundkid, line 2: Games will need to be added to email-alerts for this work.

earthboundkid: If you have the blue lena API key, you can run this to get the ID number for various lists:

```
curl --request GET \
     --url https://spotlightpa.api-us1.com/api/3/lists \
     --header "Api-Token: $BLUE_LENA" \
     --header 'accept: application/json'
```

Right now, the top list is ID 23 for Berks Ideas Festival. Someone needs to make the games list so we can grab the ID and add it to email-alerts.

earthboundkid (approved):

earthboundkid (approved):

